PR #61, opened 2018-05-25
New TCP transport implementation

rtobar:
This pull request is to include an initial version of a new TCP transport implementation. The code tries to follow closely the design of the current software stack, reusing as much as possible from the existing code, and therefore is quite compact. The changes also try to follow as closely as possible the coding style in use to be as disruptive as possible.

Two new classes have been added: the `spead2::send::tcp_stream` class takes individual SPEAD packets, and sends both the packet's size (as a 64-bit number, although it probably could do with less than that) and the packet itself through a TCP socket. On the other side the `spead2::recv::tcp_reader` class uses a double-buffering mechanism to temporarily store incomplete SPEAD packets that may have arrived while still being able to continue issuing asynchronous reads. Packet serialization/deserialization uses the existing machinery. The TCP receiver needs to be started first, as it is the entity that listens for incoming connections.

An improved version of this transport would probably benefit from *not* having to serialize/deserialize a heap into packets in the first place, saving CPU and some memory allocation. This would require some significant amount of additional work though, and would probably be a way more invasive change. Performance so far is quite good though (TCP send/recv pair run at ~30 [Gb/s] on a loopback link), making this improvement less urgent.

Integration is also provided into the `spead2_send` and `spead2_recv` command-line tools, with a new `--tcp` flag selecting this transport. The classes are not exposed through the python bindings yet though, but I would imagine this would not be terribly hard to do.

This work was part of our experimentation of using SPEAD over a 100 [Gb/s] link in Western Australia. The results of these experiments are available as an SDP memo. I'm slightly unsure how public are these memos, so I'm withholding to post it here. If you are interested in getting more details please drop me a line and I will surely send a copy across.

As a final note, I would like to congratulate the authors on the excellent quality of the existing code, which made this work fairly simple and enjoyable.


bmerry:
Thanks! TCP has been something I've been interesting in supporting for a while - if only to make unit testing more reliable. I'll try to get some time from my manager to review this.

Just based on your description: I would prefer not to have an out-of-band packet length. The protocol already contains sufficient information to determine the packet boundaries (#Items in header plus packet payload length item), and `mem_reader` already operates on a stream of concatenated packets, so I don't want to introduce a second serialisation format for stream data. I realise that it'll take a bit more work to reassemble packets when you can't look arbitrarily far ahead in the stream, but I think it's doable.

I'll try to get hold of the memo from my local contacts who have access to SKAO memos.

bmerry:
PS can you send me a bit more information about the memo (e.g. title) so that my colleague can locate it for me?

rtobar:
Thank Bruce for the very prompt reply!

I just mailed it today to the SDP administration, so I don't actually know when it will see the light of day more publicly. I will actually email you the document instead so you can have it promptly.

Re. the TCP packet sizes, it would be extremely simple not to send that information from the sender side, but I think on the receiver side the double-buffering would still be needed (if I understand the code correctly), regardless of how packet sizes are determined. Not sending the size would only complicate a bit more the receiver code (which would have to determine whether it has enough packet bytes to determine the packet size!), which cannot pass down packets to the stream until they are full.

I could be wrong, but after 9719ca9 the one Travis failure seem to be unrelated to the new code and something to do with the netcap support.

bmerry:
I've fixed the netmap problem on master by pinning the test to an older commit of netmap.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+        log_warning("Error in TCP receiver: %1%", error.message());
+
+    if (read_more)
+    {
+        enqueue_receive();
+    }
+    else
+    {
+        peer.close();
+        this->stopped();
```
bmerry: The acceptor should be closed somewhere, either here or when the peer connection is established.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+
```
bmerry: Only need one blank line here.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+        {
+            log_info("TCP reader: discarding packet received after stream stopped");
+        }
+        else
+        {
+            read_more = process_buffer(bytes_transferred);
+        }
+    }
+    else if (error != boost::asio::error::operation_aborted)
+        log_warning("Error in TCP receiver: %1%", error.message());
```
bmerry: What about end-of-file (due to the remote closing the TCP connection)? I think that probably should not log a warning.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+    std::swap(buffer, buffer2);
+    if (buffer2_bytes_avail == 0) {
+        buffer_offset = 0;
+    }
+
+}
+
+bool tcp_reader::process_buffer(const std::size_t bytes_recv)
+{
+
```
bmerry: Remove blank line.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+    // scattered through the previous and current buffer,
+    // or fully in the current buffer
+    if (buffer2_bytes_avail >= 8)
+    {
+        pkt_size = load_be<std::uint64_t>(buffer2.get() + buffer_offset);
+        buffer2_bytes_avail -= 8;
+        buffer_offset += 8;
+        bytes_avail -= 8;
+    }
+    else if (buffer2_bytes_avail > 0)
```
bmerry: Need to check that `bytes_avail > 8` as well, otherwise your read from buffer could overrun.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+        bytes_avail -= 8;
+    }
+    else if (bytes_avail >= 8)
+    {
+        pkt_size = load_be<std::uint64_t>(buffer.get() + buffer_offset);
+        buffer_offset += 8;
+        bytes_avail -= 8;
+    }
+    else
+        return false;
```
bmerry: Should `pkt_size` be set to 0 here, to indicate that the packet size isn't available yet? Otherwise I think the previous packet size might still be set, and I'm not sure if that will confuse the state machine.

include/spead2/recv_tcp.h
```diff
@@ -0,0 +1,145 @@
...
+    void accept_handler(
+        const boost::system::error_code &error);
+
+    /// Callback on completion of asynchronous receive
+    void packet_handler(
+        const boost::system::error_code &error,
+        std::size_t bytes_transferred);
+
+    /// Processes the content of the buffer
+    /// Retruns true if more reading needs to be enqueued
```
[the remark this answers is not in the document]
bmerry: It's somewhat confusing that this function returns the opposite value to process_one_packet and parse_packet, which return true if the stream stopped.

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+            finish_buffer_processing(bytes_recv, bytes_avail);
+            return true;
+        }
+    }
+
+    while (bytes_avail >= pkt_size)
+    {
+        LOG_STEP("Trying to read packet");
+        if (parse_packet(bytes_avail)) {
+            // we don't enqueue any more reads, as the heap actually finished
```
bmerry: heap -> stream

src/recv_tcp.cpp
```diff
@@ -0,0 +1,308 @@
...
+    // we couldn't consume a single byte of this->buffer, and therefore the
+    // offset refers to this->buffer2. If that's the case then we need to concatenate
+    // the contents of the two buffers onto the current buffer (doesn't matter
+    // exactly where, but we put it in position 0), and adjust the necessary bits
+    // of information
+    if (buffer2_bytes_avail > 0)
+    {
+        // we are not doing this optimally, but it's not happening that often
+        LOG_STEP("Couldn't process any received data, accumulating it into buffer");
+        std::uint8_t buf[max_size];
```
bmerry: This is potentially a large allocation, so it's not a good idea to have it on the stack (particularly since if it overflows the stack, you'll only find out under exceptional circumstances in production, rather than the first time you try it). It should be possible to shuffle the data safely in-place using `memmove`.

src/spead2_recv.cpp
```diff
@@ -46,6 +47,7 @@ struct options
     bool descriptors = false;
     bool pyspead = false;
     bool joint = false;
+    bool tcp = false;
     std::size_t packet = spead2::recv::udp_reader::default_max_size;
     std::size_t buffer = spead2::recv::udp_reader::default_buffer_size;
```
bmerry: Let's make the default depend on whether tcp was set i.e. don't set a default in the parser, and if no value was specified than manually set it to either udp_reader::default_buffer_size or tcp_reader::default_buffer_size depending on which was specified.

src/spead2_recv.cpp
```diff
@@ -214,6 +217,36 @@ void show_heap(const spead2::recv::heap &fheap, const options &opts)
...
+
+        if (wrong)
+            std::cerr << "Wrong values found in item 0x1000\n";
+        else if (!opts.quiet)
+            std::cout << "All values in item 0x1000 seem correct\n";
+
+        break;
+    }
+}
+
```
bmerry: I'm assuming this was added for testing your implementation, but it's a pretty weak test: a batch of all zeros can easily turn up even if things are wrong, and it'll pass if the item 0x1000 disappears completely. The Python bindings have lots of better unit tests, so I'd rather take this out and when I plumb in the Python API I'll test it properly from there.

src/spead2_recv.cpp
```diff
@@ -291,6 +328,15 @@ static std::unique_ptr<spead2::recv::stream> make_stream(
         }
         else
             port = *i;
+
+        if (opts.tcp) {
+            tcp::resolver resolver(thread_pool.get_io_service());
+            tcp::resolver::query query(host, port, tcp::resolver::query::address_configured);
+            tcp::endpoint endpoint = *resolver.resolve(query);
+            stream->emplace_reader<spead2::recv::tcp_reader>(endpoint, opts.packet, opts.buffer);
+            continue;
```
bmerry: I'd prefer to put the UDP code in an `else` branch rather than using `continue`.

src/spead2_recv.cpp
```diff
@@ -337,9 +383,9 @@ int main(int argc, const char **argv)
     }
     else
     {
-        if (opts.sources.size() > 1 && opts.ring)
+        if (opts.sources.size() > 1 && (opts.ring or opts.tcp))
```
bmerry: Use `||` rather than `or`.

src/spead2_recv.cpp
```diff
@@ -337,9 +383,9 @@ int main(int argc, const char **argv)
...
         {
-            std::cerr << "Multiple streams cannot be used with --ring\n";
+            std::cerr << "Multiple streams cannot be used with --ring or --tcp\n";
```
bmerry: Why don't multiple streams work with TCP?

include/spead2/send_tcp.h
```diff
@@ -0,0 +1,100 @@
...
+{
+private:
+    friend class stream_impl<tcp_stream>;
+    boost::asio::ip::tcp::socket socket;
+    boost::asio::ip::tcp::endpoint endpoint;
+    std::uint64_t packet_size;
+    boost::asio::const_buffers_1 size_buffer;
+
+    template<typename Handler>
+    void async_send_packet(const packet &pkt, Handler &&handler)
```
bmerry: This is a little large to be inlined (although once the out-of-band packet size is removed it might be okay).

include/spead2/send_tcp.h
```diff
@@ -0,0 +1,100 @@
...
+    {
+        packet_size = betoh(std::uint64_t(boost::asio::buffer_size(pkt.buffers)));
```
bmerry: Should be htobe.

include/spead2/send_tcp.h
```diff
@@ -0,0 +1,100 @@
...
+            if (ec) {
+                // not sure what to do really...
+                return;
+            }
+            boost::asio::async_write(socket, pkt.buffers, std::move(handler));
+        });
+    }
+
+public:
+    /// Socket receive buffer size, if none is explicitly passed to the constructor
```
bmerry: Send buffer, not receive buffer (I'll fix it in send_udp.h).

src/send_tcp.cpp
```diff
@@ -0,0 +1,101 @@
...
+    std::size_t buffer_size)
+    : stream_impl<tcp_stream>(std::move(io_service), config),
+    socket(std::move(socket)), endpoint(endpoint),
+    packet_size(0), size_buffer(boost::asio::const_buffer(&packet_size, sizeof(packet_size)))
+{
+    if (!socket.is_open())
+    {
+        if (!local_endpoint.address().is_unspecified())
+            this->socket.bind(local_endpoint);
+        this->socket.connect(endpoint);
```
bmerry: This is a blocking operation. That could be problematic for using this in an event-driven program (e.g. if it were connected to Python asyncio), although I'm not sure of the best solution. One option would be to have the constructor take a callback handler and pass it to `async_connect`.

src/send_tcp.cpp
```diff
@@ -0,0 +1,101 @@
...
+            // Linux silently clips to the maximum allowed size
+            boost::asio::socket_base::send_buffer_size actual;
+            this->socket.get_option(actual);
+            if (std::size_t(actual.value()) < buffer_size)
+            {
+                log_warning("requested socket buffer size %d but only received %d: refer to documentation for details on increasing buffer size",
+                            buffer_size, actual.value());
+            }
+        }
+    }
```
bmerry: This code is a duplicate - let's move it somewhere common - maybe send_stream.cpp, or else make a `common_socket.cpp` that has all the socket option manipulation wrappers.

src/spead2_send.cpp
```diff
@@ -93,6 +97,8 @@ static options parse_args(int argc, const char **argv)
         ("heaps", make_opt(opts.heaps), "Number of data heaps to send (-1=infinite)")
         ("pyspead", make_opt(opts.pyspead), "Be bug-compatible with PySPEAD")
         ("addr-bits", make_opt(opts.addr_bits), "Heap address bits")
+        ("tcp", make_opt(opts.use_tcp), "Use TCP instead than UDP")
+        ("tcp-local", make_opt(opts.tcp_local), "Local address to bind TCP sockets to (none by default)")
```
bmerry: Binding to a particular local address could be just as useful for UDP. Would you be interested in implementing that? At the minimum can you call this option `"bind"` so that it's not TCP-specific (the name matches an old name in spead2_recv that let the receiver bind to a particular address).

src/spead2_send.cpp
```diff
@@ -243,9 +249,6 @@ int main(int argc, const char **argv)
     options opts = parse_args(argc, argv);
 
     spead2::thread_pool thread_pool(opts.threads);
-    udp::resolver resolver(thread_pool.get_io_service());
-    udp::resolver::query query(opts.host, opts.port);
-    auto it = resolver.resolve(query);
```
bmerry: Removing this is going to make the SPEAD2_USE_IBV codepath fail to compile. Also if `--ibv` and `--tcp` are both given the tool should report an error to the user.

rtobar:
Thanks a lot for the very thorough review. I've already been addressing most of the comments you gave me on the receiver side, and I will try to go through the new ones hopefully tomorrow. I will post the new patches once I have them for you to review them again.

bmerry:
Thanks. It looks like you've done a lot of work - I'll get around to reviewing it when I can.

Re the slow-down with the single buffer: is that something I'd be able to reproduce on loopback?

rtobar:
I think so. That is where I tested at least, using IPv4. I don't have a clear explanation for it, but hopefully will be obvious to you once you go through the code.

src/spead2_send.cpp
```diff
@@ -98,7 +98,7 @@ static options parse_args(int argc, const char **argv)
         ("pyspead", make_opt(opts.pyspead), "Be bug-compatible with PySPEAD")
         ("addr-bits", make_opt(opts.addr_bits), "Heap address bits")
         ("tcp", make_opt(opts.tcp), "Use TCP instead than UDP")
-        ("tcp-local", make_opt(opts.tcp_local), "Local address to bind TCP sockets to (none by default)")
+        ("bind", make_opt(opts.bind), "Local address to bind sockets to")
```
bmerry: Can you add a check that if `--bind` is used without `--tcp` then it will fail with an error message indicating the `--bind` is not yet supported for UDP.

src/spead2_recv.cpp
```diff
@@ -49,7 +49,7 @@ struct options
     bool joint = false;
     bool tcp = false;
     std::size_t packet = spead2::recv::udp_reader::default_max_size;
-    std::size_t buffer = spead2::recv::udp_reader::default_buffer_size;
+    std::size_t buffer = 0;
```
bmerry: This is a bit slippery because set_socket_buffer_size treats 0 as "leave the OS default", and I'd like to reflect that behaviour in the command-line tools. When the option is defined, you should be able to define it without a default at all (you'll need to take out the `make_opt` and construct the option directly), and then query `vm` to find out whether the option was passed.

bmerry:
I haven't been able to reproduce a slowdown with the tcp_single_buffer branch using loopback. I'm testing with
```
src/spead2_recv --quiet --tcp localhost:8888
src/spead2_send --tcp --packet 9200 --heaps 5000 localhost 8888
```
and the sender is reporting 16-17 Gb/s with both branches. The tcp_single_buffer branch is maybe a touch slower, but it's in the noise, whereas it sounds like you've got a case where it is very clearly slower. Can you send me the arguments you're using on each end?

The code looks sensible, and much easier to follow. The only thought I had is that if you wait until the head goes past half-way, rather than the tail, then it will be safe to use memcpy rather than memmove - not that I expect any significant performance difference between them, particularly for a copy from a higher to a lower address. Another option would be to wait until the remaining space is less than max_size, which would reduce the number of copies but possibly increase the number of syscalls.

include/spead2/common_socket.h
```diff
@@ -32,16 +32,24 @@ namespace spead2
 {
 
 /**
- * Tries to sets the socket's buffer size option, warns if it cannot be done
+ * Tries to sets the socket's send buffer size option, warns if it cannot be done.
+ * If @p buffer_size is zero the socket is left unchanged.
```
bmerry: I use `@a` rather than `@p` for function arguments.

include/spead2/common_socket.h
```diff
@@ -32,16 +32,24 @@ namespace spead2
...
  * @param buffer_size The buffer size
  */
 template <typename SocketType>
-void set_socket_buffer_size(SocketType &socket, std::size_t buffer_size);
+void set_socket_send_buffer_size(SocketType &socket, std::size_t buffer_size);
 
-extern template void set_socket_buffer_size<boost::asio::ip::tcp::socket>(boost::asio::ip::tcp::socket &socket, std::size_t buffer_size);
-extern template void set_socket_buffer_size<boost::asio::ip::udp::socket>(boost::asio::ip::udp::socket &socket, std::size_t buffer_size);
+/**
+ * Tries to sets the socket's receive buffer size option, warns if it cannot be done.
```
bmerry: sets -> set, and an "and" before "warns".

rtobar:
I just pushed the docstring changes.

I had realized that the OSX failure was intermittent. I guess this is what you were referring to when you said you were considering TCP at least for having more reliable unit tests?

Re. the performance difference, I'm using a bigger packet size in the `lo` interface, which reports an mtu of 65536:

```
sudo nice -n 0 src/spead2_send --tcp 127.0.0.1 5001 --heaps 10000 --packet 64500
sudo nice -n 0 src/spead2_recv --tcp 127.0.0.1:5001 --packet 64500 --quiet
```

bmerry:
> I had realized that the OSX failure was intermittent. I guess this is what you were referring to when you said you were considering TCP at least for having more reliable unit tests?

Exactly.

> Re. the performance difference, I'm using a bigger packet size in the lo interface, which reports an mtu of 65536:

Okay, with that I can more reliably see a (small) performance difference: 45.3 Gb/s with single buffer, 46.4-47.4 Gb/s with double buffer.

bmerry:
<s>Okay, with that I can more reliably see a (small) performance difference: 45.3 Gb/s with single buffer, 46.4-47.4 Gb/s with double buffer.</s>

Correction, that was with a build where I'd experimented with changing the code. With your branch as-is, I get 46.2-46.8 Gb/s with the commands you specified, and if I pin the sender and receiver to separate cores I get up to 48.0 Gb/s. So I'm not seeing the 25% swings you are.

rtobar:
Today I run some more tests in machines other than my personal computer, and I confirmed that the single buffer version of the code performs similarly to the double buffer version. The fastest speed I have achieved is < 40 [Gb/s] though, so I wonder whether I am pinning the processes correctly.

If you agree, I will put the single buffer version of the code on top of the `tcp` branch and remove the `tcp_single_buffer` branch. I will also go ahead with your suggestion regarding checking the head instead of the tail to perform a `memcpy` instead of a `memmove`. I should also be able to also implement a sensible packet size check on top of this.

bmerry:
That all sounds good. For what it's worth, my results are on a i7-7820HK (Kaby Lake). It's a laptop part, but they seem to have done some magic with the memory interface and I've seen it significantly outperform Broadwell Xeons with similar clock speeds and much bigger caches on single-threaded memory-bound tasks.

bmerry:
Thanks for the updates. I'm a bit snowed under this week, but I'll take a look when I can.

rtobar:
No hurry, I will be actually away for the next two weeks so I won't be able to proceed any further during that time.

bmerry:
The latest batch of changes seems reasonable. Apart from the out-of-band passing of the packet size, I think there are still one or two unaddressed comments, particularly regarding the blocking connect call.

Regarding getting the packet size from inside the packet: I'm pretty rusty on the interfaces in my packet parsing code, but let me know if I can help with extending it so that it can take a (possibly partial) packet and indicate whether more data is needed to complete the packet. I guess a particularly nasty case is when there are so many items that the packet size doesn't appear in the first `max_packet` bytes.

rtobar:
Here is a new set of patches. They remove the out-of-band size transmission, which was one of the main points to be addressed. There was a big inlined piece of code that you had commented on the tcp sender side which is now gone as well. I think with these changes the only main one is the callback-providing constructor for the sender, which I will provide shortly.

Apart from these changes, I also reverted the setting of the receive buffer size to be applied to the acceptor and not the data object. Testing on our long, wide link showed that setting the receive buffer size on the data socket did not work as expected, and that setting it on the acceptor did. Please see the commit messages for more details.

bmerry:
Thanks for the updates. I'm going to be away for the next week so will have to check them out when I get back. Good catch on the socket buffer size issue.

I've been adding some code of my own to spead2, but hopefully it won't have more than trivial conflicts.

rtobar:
And a final set of changes to support asynchronous TCP connects.

rtobar:
Just a couple more of updates, mostly cleaning up after building with `-Wall` (didn't realize it wasn't on by default until now).

The only important bit is the adding back of explicit definitions for constexpr static members, which is unnecessary with C++17, but still very much so with 11. I hadn't stumble upon these errors until now because the `spead2_recv` program still defaults the `--packet` option to `udp_reader::default_max_size`, although to be correct it should use something similar as what we have for the `--buffer` option (using either `udp_reader`'s or `tcp_reader`'s member as default depending on `--tcp`). A different option which I also considered at some point, is to have an `ip_base_reader` class defining a `default_max_size`, or even have this member reside on the base `reader` class. I leave this decision to you though, as you probably have a much better feeling on how to structure this.

include/spead2/send_tcp.h
```diff
@@ -51,36 +52,68 @@ class tcp_stream : public stream_impl<tcp_stream>
     static constexpr std::size_t default_buffer_size = 512 * 1024;
 
     /// Constructor
+    template <typename ConnectHandler>
```
bmerry: I generally don't put a space between `template` and `<`.

include/spead2/send_tcp.h
```diff
@@ -51,36 +52,68 @@ class tcp_stream : public stream_impl<tcp_stream>
...
         const boost::asio::ip::tcp::endpoint &endpoint,
         const boost::asio::ip::tcp::endpoint &local_endpoint = boost::asio::ip::tcp::endpoint(),
         const stream_config &config = stream_config(),
-        std::size_t buffer_size = default_buffer_size);
+        std::size_t buffer_size = default_buffer_size)
+        : stream_impl<tcp_stream>(std::move(io_service), config),
+          socket(std::move(socket)), endpoint(endpoint)
+    {
+        if (&get_io_service() != &this->socket.get_io_service())
+            throw std::invalid_argument("I/O service does not match the socket's I/O service");
```
bmerry: Add `<exception>` to the includes to ensure invalid_argument is defined.

include/spead2/send_tcp.h
```diff
@@ -51,36 +52,68 @@ class tcp_stream : public stream_impl<tcp_stream>
...
+        set_socket_send_buffer_size(this->socket, buffer_size);
+        if (!socket.is_open())
```
bmerry: Should be `this->socket`.

include/spead2/send_tcp.h
```diff
@@ -51,36 +52,68 @@ class tcp_stream : public stream_impl<tcp_stream>
...
+        {
+            if (!local_endpoint.address().is_unspecified())
+                this->socket.bind(local_endpoint);
+            this->socket.async_connect(endpoint, connect_handler);
+        }
+    }
+
```
bmerry: Can eliminate these two blank lines.

src/spead2_send.cpp
```diff
@@ -277,8 +277,17 @@ int main(int argc, const char **argv)
         tcp::endpoint local_endpoint;
         if (!opts.bind.empty())
             local_endpoint = tcp::endpoint(boost::asio::ip::address::from_string(opts.bind), 0);
+
+        auto promise = std::make_shared<std::promise<void>>();
```
bmerry: Is it necessary for this to be a shared ptr, rather than just a local that is captured by reference?

bmerry:
I just realised there was one batch of updates I skipped over (to do with eliminating the out-of-band packet size transmission); will take a look tomorrow.

Do you recall why the tcp_reader constructor that takes an acceptor socket expects it to be pre-bound, while the udp_reader constructor expects the socket *not* to be bound and binds it? The TCP version possibly makes more sense (gives maximum control to the user), but the inconsistency bugs me.

bmerry:
I would like to keep the constructor that takes both an `io_service_ref` and a socket. An `io_service_ref` can hold a `shared_ptr` to a thread_pool, and thus allows constructing a stream with a thread pool in a way that the stream holds onto the thread pool without the user needing to keep their own ref alive. I will probably need that for the Python bindings.

rtobar:
OK, I'll add that back, the use case you describe wasn't obvious to me.

Regarding your earlier question about the constructor on `tcp_read` with an existing acceptor as a parameter: I did not have a particular reason on why I wrote it assuming it was already bound. I understand that this The question is probably broader though: there are now quite a few constructors taking existing sockets as parameters, and I guess you

bmerry:
Looks like your last comment got truncated.

src/recv_packet.cpp
```diff
@@ -45,33 +45,68 @@ static inline T extract_bits(T value, int first, int cnt)
     return (value >> first) & ((T(1) << cnt) - 1);
 }
 
-std::size_t decode_packet(packet_header &out, const uint8_t *data, std::size_t max_size)
+static
+bool decode_spead_header(const uint8_t *data, int &heap_address_bits, int &n_items)
```
bmerry: I think just call this `decode_header`, and put the `static` on the same line as the rest of the signature.

src/recv_packet.cpp
```diff
@@ -45,33 +45,68 @@ static inline T extract_bits(T value, int first, int cnt)
...
+
+s_item_pointer_t get_packet_size(const uint8_t *data, std::size_t length)
+{
+    if (length < 8)
+        return 0;
+    int heap_address_bits, n_items;
+    if (!decode_spead_header(data, heap_address_bits, n_items))
+        return -1;
+    if (std::size_t(n_items) * sizeof(item_pointer_t) + 8 > length)
         return 0;
```
bmerry: In theory one could look through the subset of items that do fit (typically payload length will be before user-defined items), but it's probably not worth worrying about that corner case.

include/spead2/recv_tcp.h
```diff
@@ -88,6 +88,8 @@ class tcp_reader : public udp_reader_base
     bool skip_bytes();
 
 public:
+    /// Maximum packet size, if none is explicitly passed to the constructor
+    static constexpr std::size_t default_max_size = 9200;
```
bmerry: For TCP I think it may be better to use a bigger max size e.g. 65536. For UDP this value was reasonable as typically Ethernet won't support larger MTUs and one wants to avoid fragmentation, but as you've noticed TCP benefits from larger packets.

src/recv_tcp.cpp
```diff
@@ -145,10 +145,19 @@ bool tcp_reader::parse_packet_size()
...
+    auto s_pkt_size = get_packet_size(head, tail - head);
+    if (s_pkt_size == -1)
+    {
+        /* We only skip the first 8 bytes (i.e., the SPEAD header) hoping that
+         * a new packet will appear later with a correct header later.
+         */
+        log_info("discarding packet due to invalid header");
+        head += 8;
+        return false;
+    }
```
bmerry: I think at some stage the right approach may be just to shut down the stream. The probability of finding something in the data that looks enough like a header is fairly high and will probably lead to a storm of errors. Let's leave it for now though and fix it if it becomes an issue.

src/recv_tcp.cpp
```diff
@@ -145,10 +145,19 @@ bool tcp_reader::parse_packet_size()
...
+    else if (s_pkt_size == 0)
```
bmerry: Can you maybe add a comment here and on the previous if statement to remind the reader what the magic values -1 and 0 mean.

src/recv_tcp.cpp
```diff
@@ -145,10 +145,19 @@ bool tcp_reader::parse_packet_size()
...
         return true;
```
bmerry: What happens if you fill up the buffer and there still isn't enough data to determine the packet size (e.g. if n_items is very large)?

include/spead2/recv_tcp.h
```diff
@@ -89,11 +91,9 @@ class tcp_reader : public reader
 
 public:
     /// Maximum packet size, if none is explicitly passed to the constructor
-    static constexpr std::size_t default_max_size = 9200;
+    static constexpr std::size_t default_max_size = 65536;
     /// Socket receive buffer size, if none is explicitly passed to the constructor
-    static constexpr std::size_t default_buffer_size = 8 * 1024 * 1024;
-    /// Number of packets to hold on each buffer for asynchronous receive
-    static constexpr std::size_t pkts_per_buffer = 64;
+    static constexpr std::size_t default_buffer_size = 208 * 1024;
```
bmerry: Is this magic number a result of testing?

src/send_tcp.cpp
```diff
@@ -29,25 +29,25 @@ namespace send
...
+} // namespace detail
 
 constexpr std::size_t tcp_stream::default_buffer_size;
 
 tcp_stream::tcp_stream(
     io_service_ref io_service,
-    boost::asio::ip::tcp::socket &&socket,
-    const stream_config &config)
+    const stream_config &config,
+    boost::asio::ip::tcp::socket &&socket)
```
bmerry: I preferred this the way it was, for consistency with udp_stream where the socket is before the config (in general I think I put `config` after addressing-related parameters, and before optional tuning parameters like buffer_size; and where it makes sense I default it to `stream_config()`).

src/spead2_send.cpp
```diff
@@ -279,7 +279,7 @@ int main(int argc, const char **argv)
             local_endpoint = tcp::endpoint(boost::asio::ip::address::from_string(opts.bind), 0);
 
         auto promise = std::promise<void>();
-        auto connect_handler = [&promise](boost::system::error_code e) {
+        auto connect_handler = [&promise] (boost::system::error_code e) {
```
bmerry: `{` on a new line please.

include/spead2/recv_tcp.h
rtobar: Yes, that's the default `sys.net.core.rmem_max` value for Linux, which is what we have been using for the `lo`-based tests

bmerry:
Thanks again for all the work on this. I've lost track a bit of all the things I asked you to do - is there anything outstanding that you know of? The one thing that's still in my mind is the question of whether custom sockets should be bound by the caller or the callee, but I'll look into that a bit closer to check what I've already done with UDP and maybe fix it up after merging.

I've already started implementing Python bindings and connecting them up to the unit tests, which seems to be working so far.

rtobar:
The main original points you pointed out (out of band size, double-buffering, async connect) have for sure been addressed, and I have tried to be thorough on addressing the other change requests you have been pointing out. Going through the list of "outstanding requests" produced by GitHub in this page there is nothing obvious left unaddressed. I reckon there still might be one or two styling issues somewhere, but I believe the main functionality is in place. The python bindings might bring up a few more issues, but hopefully it won't be too much.

The only thing I have in mind is the fact that `spead2_send` still defaults the packet size to a default value set on the `spead_config` class. I could have broken down that and put a new member on `udp_stream` and `tcp_stream`, but the python bindings seem to be using this value too, so I decided not to step into that one.

bmerry:
Great. In that case I'm going to update my branch with your latest changes and then get down to serious work on the Python bindings. I'll leave this PR open so that if you think of anything that got missed you can still push it.

bmerry:
Just to give you an update on progress: I'm nearly ready to get this merged, but I'm currently running into problems on the Travis Mac builds due to pybind/pybind11#1473.

rtobar:
Thanks for the update Bruce, I'm really looking forward to seeing the python support exposed.

On a different topic, I realized that there is no timeout support yet in this initial implementation. I don't think I have the time really to go through it at the moment, but just wanted to note it as possible future work.

bmerry:
Timeouts is an interesting point. Which operations were you thinking of timing out? The user could probably already do some of it by setting their own socket timeouts on custom sockets, but one would want to check that the timeout errors that come back are handled gracefully. I think can also time out connecting by having ones own timer and then deleting the tcp_stream.

But as you say, let's leave that for a later version.

rtobar:
I was thinking on timing out both on the sender and the receiver sides. Even though the transmission is asynchronous, a final operation depending on the stream transmission finishing would not get executed if a sender/receiver hangs.

I quickly checked and asio doesn't seem to support timeouts via socket options, so the implementation would have to be using a timer within the spead2 classes, and the timeout would then be a parameter given by the user (or one could read the SO_RCVTIMEO/SO_SENDTIMEO options set on the socket).

bmerry:
Something I've been meaning to get around to for a while is the option to abort transmission (currently the destructor gracefully waits for transmission to complete). I prefer to provide abort functionality than timeouts because it is more general and can be used to implement abort conditions other than "in X seconds from now".

I've managed to work around the hang in the unit tests, so I'm going to merge my `tcp` branch to master and close this PR. I still want to add some more unit tests to cover the various special cases and error paths before I cut a release.

PR #73, opened 2019-09-16
Remove support for Trollius

bmerry:
Trollius is no longer maintained, and I'm not aware of the support for it being used any more (there are still a few Python 2 users, but they're using the synchronous API).


spead2/send/asyncio.py
```diff
@@ -1,3 +1,5 @@
+# Generated by bootstrap.sh from spead2/send/asyncio.py. Do not edit.
```
sratcliffe: Isn't this file spead2/send/asyncio.py :)

spead2/send/asyncio.py
```diff
@@ -212,9 +212,12 @@ def callback(arg):
...
             thread (if `comp_vector` is negative).
-        loop : :py:class:`trollius.BaseEventLoop`, optional
-            Event loop to use (defaults to ``trollius.get_event_loop()``)
+        loop : :py:class:`asyncio.BaseEventLoop`, optional
+            Event loop to use (defaults to ``asyncio.get_event_loop()``)
         """
 
 except ImportError:
     pass
+# Monkey-patch asyncio.ensure_future for Python < 3.4.4.
```
sratcliffe: Didn't we drop support for 3.4.x already ?

spead2/send/asyncio.py
bmerry: Thanks for catching that - this file was originally generated but is now canonical.

spead2/send/asyncio.py
bmerry: Thanks, I missed stripping that bit out.

PR #74, opened 2019-09-22
Remove Python 2 support

bmerry:


doc/introduction.rst
```diff
@@ -12,7 +12,7 @@ PySPEAD_ implementation, spead2:
 - cleanly supports several SPEAD flavours (e.g. 64-40 and 64-48) in one
   module, with the receiver adapting to the flavour used by the sender;
 - supports Python 3;
-- supports asynchronous operation, using asyncio_ (Python 3 only).
+- supports asynchronous operation, using asyncio_.
```
sratcliffe: Should we mention the 3.5 and above in the supports line above?

bmerry: Python 3.4 is already past end-of-life and it's encoded into setup.py, so I'm not bothered.

PR #77, opened 2019-10-07
Reduce number of cyclic refs

bmerry:


bmerry:
@sjperkins this may interest you.

spead2/send/__init__.py
```diff
@@ -110,7 +112,8 @@ def add_to_heap(self, heap, descriptors='stale', data='stale'):
             raise ValueError("descriptors must be one of 'stale', 'all', 'none'")
         if data not in ['stale', 'all', 'none']:
             raise ValueError("data must be one of 'stale', 'all', 'none'")
-        for item in self._item_group.values():
+        item_group = self._item_group if self._item_group is not None else self
+        for item in item_group.values():
```
sratcliffe: I am almost certainly just missing something here, but doesn't this potentially end up with a .values() called on a HeapGenerator object, which doesn't seem to provide values() ?

bmerry: This happens when using the `spead2.send.ItemGroup` class, which inherits from both `spead2.ItemGroup` and `spead2.send.HeapGenerator` and hence has the methods of both. It cancels out the transformation at line 56, but avoids the object storing a reference to `self`.

PR #79, opened 2019-10-11
Fix a race condition in TCP accept_handler

### LINKED ISSUE #78: Race condition in TCP handling
The `accept_handler` manipulates the sockets without hold the stream's `queue_mutex`. This can cause problems such as the call to `close` failing with EBADF.

bmerry:
It was manipulating the sockets without the appropriate mutex. Apart
from generally being undefined behaviour, it could manifest as trying to
double-close the socket leading to an EBADF exception, killing the
io_service thread.

Closes #78.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+.. rubric:: Development version
+
+- Fix race condition in TCP receiver (#78)
+
 .. rubric:: 2.0.0
 
 - Drop support for Python 2.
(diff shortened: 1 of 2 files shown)


bmerry:
@sratcliffe this is the root cause of the issues I was having with katsdpbfingest.

@rtobar not sure if you're still using SPEAD over TCP/IP, but this fix may be of interest to you.

rtobar:
@bmerry yes, we still use it, but never seen that particular problem. Thanks for the notice and the fix!

PR #81, opened 2019-11-08
Allow UDP destination to be specified per-heap

bmerry:
I've marked this API as experimental, because the change has a few percent performance regression even when it's not being used, and because I have some ideas for a more generic revamp of the send API (which would allow multiple heaps to be queued at once with their packets interleaved to help switch load balancing).

There is unfortunately a lot of code refactoring because a lot of definitions that were in a base class (`stream_impl_base`) could no longer be shared, so I merged the base class into the derived class.


bmerry:
The more I think about it the less happy I am with putting this into a release. It's only half a solution to a problem: it lets one multiplex several logical flows over one transport, but it sends heaps serially rather than interleaving them (which would be required for load balancing on our switch). I'm going to close the PR and come up with a better solution, which probably involves refactoring the stream classes into high-level flows and low-level transports.

PR #84, opened 2020-02-11
Support unicast in ibverbs receiver paths

bmerry:
This includes
- library APIs
- spead2_recv and spead_recv.py test programs
- spead2_bench and spead2_bench.py benchmark programs
- mcdump
There are also a few tweaks to interpretation of arguments in the non-verbs code-paths for consistency.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -15,7 +15,7 @@ Changelog
 - Fix occasional crash when using thread pool with more than one thread
   together with ibverbs.
 - Fix bug in mcdump causing it to hang if the arguments couldn't be parsed
-  and capturing to file.
+  (only happened when capturing to file).
 
 .. rubric:: 2.0.2
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -13,6 +13,8 @@ Changelog
 - Show the values of immediate items in :program:`spead2_recv`.
 - Fix occasional crash when using thread pool with more than one thread
   together with ibverbs.
+- Fix bug in mcdump causing it to hang if the arguments couldn't be parsed
+  and capturing to file.
 
 .. rubric:: 2.0.2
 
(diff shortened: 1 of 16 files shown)


doc/changelog.rst
```diff
@@ -3,10 +3,19 @@ Changelog
...
+  number.
+- Support unicast addresses in a few APIs that previously only accepted
+  multicast addresses; in most cases the unicast address must match the
+  interface address.
 - Add missing ``<map>`` include to ``<spead2/recv_heap.h>``.
 - Show the values of immediate items in :program:`spead2_recv`.
 - Fix occasional crash when using thread pool with more than one thread
   together with ibverbs.
+- Fix bug in mcdump causing it to hang if the arguments couldn't be parsed
+  and capturing to file.
```
sratcliffe: Couldn't parse this - are we saying that arguments handling is fixed; and we can now capture to a file ?

doc/py-ibverbs.rst
```diff
@@ -8,9 +8,9 @@ of libibverbs.
 
 There are a number of limitations in the current implementation:
 
- - Only IPv4 is supported
- - VLAN tagging, IP optional headers, and IP fragmentation are not supported
- - Only multicast is supported.
+ - Only IPv4 is supported.
```
sratcliffe: There is a bunch of IPv6 support in the recv libraries - is this statement specific to using verbs ?

doc/changelog.rst
bmerry: It was only broken when capturing to file. I agree it's not the clearest wording - I'll try to improve it.

doc/py-ibverbs.rst
bmerry: Correct.

doc/changelog.rst
bmerry: Does it look better now?

PR #85, opened 2020-02-11
Fix Travis MacOS builds

bmerry:


.travis/install.sh
```diff
@@ -12,9 +12,8 @@ echo "CXX = ${CXX:=g++}"
 
 if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     if [[ "$PYTHON" == "python3" ]]; then
-        brew update
-        brew upgrade python
-        pyvenv venv
+        python3 --version
```
ludwigschwardt: Did you use brew just for python3?

bmerry: I think so.

PR #91, opened 2020-05-18
Forcibly undefined ibverbs functions after including the header

### LINKED ISSUE #90: Compilation fails with MLNX_OFED 5.0
This is reported by @ajameson:
```
libspead2.a(common_ibv.o): In function spead2::ibv_mr_t::ibv_mr_t(spead2::ibv_pd_t const&, void*, unsigned long, int)': common_ibv.cpp:(.text+0x828): undefined reference to ibv_reg_mr'
```
The problem is that RDMA-Core now defines `ibv_reg_mr` as a [macro](https://github.com/linux-rdma/rdma-core/blob/7af01c79e00555207dee6132d72e7bfc1bb5485e/libibverbs/verbs.h#L2426), which prevents the loader from functioning correctly.

bmerry:
This should fix #90.

### DIFF
--- a/gen/gen_ibv_loader.py
+++ b/gen/gen_ibv_loader.py
@@ -132,6 +132,10 @@ HEADER = PREFIX + '''\
 #include <infiniband/verbs.h>
 #include <rdma/rdma_cma.h>
 
+{% for node in nodes -%}
+#undef {{ node.name }}
+{% endfor %}
+
 namespace spead2
 {
 


bmerry:
@ajameson can you let me know if this works for you? Note that you'll need to run `./bootstrap.sh` after applying the patch to update the generated code.

bmerry:
@ajameson were you able to test this against MLNX-OFED 5.0?

ajameson:
@bmerry, I can confirm that this commit fixes the issue with MLNX-OFED 5.0 under CentOS 8

bmerry:
Thanks @ajameson. I'll merge this and release it. You might still find that 4.x performs better with spead2 because Mellanox have deprecated their previous experimental and poorly documented APIs for multi-packet receive queues and introduced a different and still poorly documented mechanism, which I'll need to figure out to support in spead2.

PR #93, opened 2020-09-10
Prepare 3.0.0b1 release

bmerry:
From the changelog:

The :doc:`ibverbs <py-ibverbs>` acceleration has been substantially modified to use a
newer version of rdma-core. It will no longer compile against versions of
MLNX-OFED prior to 5.0. Compiled code (such as Python wheels) will still run
against old versions of MLNX-OFED, but extension features such as multi-packet
receive queues and packet timestamps will not work. It is recommended that if
you are using ibverbs acceleration with older MLNX-OFED drivers that you stick
with spead2 2.x until you're able to upgrade the drivers and spead2
simultaneously.

Other changes:

- Support hardware send rate limiting when using ibverbs.
- Discover libibverbs and pcap using pkg-config where possible.
- Make :program:`configure` print out the configuration that will be compiled.
- Update the Python wheels to use manylinux2014. This uses a newer compiler
  (potentially giving better performance) and supports :c:func:`sendmmsg`.
- Add wheels for Python 3.9.
- A number of deprecated functions have been removed.
- Avoid ibverbs code creating a send queue for receiver or vice versa.
- Rename ``slave`` option to :program:`spead2_bench` to ``agent``.


src/mcdump.cpp
```diff
@@ -900,30 +850,26 @@ void capture::network_thread()
...
-                    if (timestamp_support && (wc[i].exp_wc_flags & IBV_EXP_WC_WITH_TIMESTAMP))
+                    std::uint32_t byte_len = ibv_wc_read_byte_len(cq.get());
+                    record.incl_len = byte_len;
+                    record.orig_len = byte_len;
+                    if (timestamp_support)
                     {
-                        std::uint64_t ticks = wc[i].timestamp - timestamp_subtract;
-                        std::uint64_t ns = std::uint64_t(ticks * hca_ns_per_clock);
-                        ns += timestamp_add;
+                        std::uint64_t ns = ibv_wc_read_completion_wallclock_ns(cq.get());
```
gcallanan, line 872: This is quite nifty. It does the conversion to nanoseconds for you. 

I imagine that this is not something that you plan to add to the packet metadata in the chunks/packets/slots produced by udp_ibv_reader?(not sure what the correct terminology is here).

mcdump is probably the main use case for these timestamps, but it might be useful to have a timestamp of when a chunk received data. Just a thought. No hard requirements on CBFs side.

gcallanan (commented):
I have read through the changes and I think I am clear on what is going on. So basically, NIC timestamp monitoring and rate limiting are now part of the standard ibverbs api. The multi-packet receive queues implemented using ibverbs_exp are no longer as well supported - they require the mlx5dv.h file to implement. 

It looks like mcdump has been overhauled to make use of the new inbuilt timestamp monitoring functionality and the recv_udp_ibv_mprq.h file has been completely rewritten to implement mprq manually. The mprq functionality is not very clear to me. (I have made a separate comment for this.)

You mentioned in your presentation today that the CAP_NET_RAW capability needs to be given to the executable running ibverbs, will this be listed in your SPEAD2 documentation on readthedocs.io?

SPEAD2 is quite a beast with lots of moving parts. I have just looked over the ibverbs specific parts. The rest I have just left alone, to be revisited another day.

src/recv_udp_ibv_mprq.cpp
```diff
@@ -42,44 +42,33 @@ namespace recv
 
 static ibv_qp_t create_qp(const rdma_cm_id_t &cm_id,
```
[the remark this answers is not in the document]
bmerry: They're to do with receive-side scaling: you can attach multiple WQs (work queues) to a QP (queue pair) via a ibv_rwq_ind_table, and the hardware will load-balance packets across the work queues based on hashing header fields like src/dst IP address/port. The idea is then you can service each work queue with a separate thread.

In our case we're just attaching one WQ, so the hashing is kinda irrelevant, but you still have to go this roundabout route because it's the WQ that has an interfere to get a multi-packet version.

src/mcdump.cpp
bmerry: I don't have any plans for it. I haven't looked at exactly what the overhead is, but the driver does some work to sync the host and NIC clocks so I don't think it's free.

bmerry:
I see coveralls thinks coverage decreased a lot. I'll need to look into that - probably the optional features not being picked up on the Travis build machine.

bmerry:
> I see coveralls thinks coverage decreased a lot. I'll need to look into that

Mostly it seems that (for reasons I don't follow) the coverage collection was no longer counting results from both sets of tests (Python and C++). There are still differences but it seems to be due to the change in compiler version (due to using the Focal image to get support for all the optional features), which seems to map basic blocks to source lines differently and hence causes some lines to be marked as uncovered even though at a source level they clearly are (e.g. a `while (true)` line is marked non-covered even though the line inside the loop are covered).

bmerry:
@gcallanan I wasn't quite sure if you were planning to add more comments later, but I'm going to merge so long so I can make an alpha release. You're still welcome to add comments post-merge.

PR #94, opened 2020-09-21
Receive-in-order mode

bmerry:
By default, assume packets within a heap arrive in order. This should reduce the cost of dealing with packet loss, which previously needed to keep a list of all the contiguous ranges of a heap that had arrived so that duplicated could be detected. A large part of the changes are not related to this, but rather some refactoring. Start with the changes to doc/changelog.rst, doc/migrate-3.rst and doc/py-recv.rst for more detailed explanations.

I'd hoped that these changes would also improve general performance by eliminating a lock for accessing the config (no longer needed now that the config is immutable) and bypassing the complicated logic around the `std::map` for tracking received data, but if so it's down in the noise.


src/recv_live_heap.cpp
```diff
@@ -179,8 +180,25 @@ bool live_heap::add_packet(const packet_header &packet,
     }
 
     // Packet seems sane, check if we've already seen it, and if not, insert it
-    bool new_packet = add_payload_range(packet.payload_offset,
-                                        packet.payload_offset + packet.payload_length);
+    bool new_packet;
+    if (allow_out_of_order)
```
gcallanan, line 184: I would like to clarify what happens to a heap by default when a packet is rejected and allow_out_of_order is false. Is the heap dropped? (Although this whole series of events is unlikely to happen)

gcallanan (commented):
There is quite a lot going on here, not all of it is clear to me. I have looked over the main changes and most of it seems to be in order. In summary, the main changes introduced in this pull request are:
1. Added logic to simplify heap assembly when packets are expected to be received in order.
2. Streamlined the stream configuration option to two objects: a StreamConfig() and RingStreamConfig() object

I also spent some time looking at the examples and the documentation. I have some thoughts on what could be added there. They are not related to the changes made in this review, so consider them as suggestions.
1. When reading the documentation for the first time, it is difficult to figure out how one would start a SPEAD2 project. The documentation contains a detailed API, but it requires a bit of legwork to figure out how all the different bits fit together(eg: You know you have a memory pool and a thread pool, but then you need to go over the rest of the docs to figure out where they fit in). The examples in spead2/examples serve to show how all the parts interact, but I could not find any reference to these examples in the docs. If you install spead2 using "pip install", then it is quite likely that you wont even know that the examples exist. To make this more accessible to first time users, I would recommend explicitly mentioning the examples somewhere in the documentation.
2. There is a test_send_asyncio.py script in the examples. Is it worth having a test_receive_asyncio.py script? In recv.py in katfgpu there is a AsyncRingbuffer class. Was this specific to katfgpu or would this be something someone using standard spead2 would need to consider?

src/recv_live_heap.cpp
bmerry: I consider them to be independent events. Some set of packets will be accepted, and if those packets form a complete heap then all is well. If packets are rejected due to reordering (rather than, for example, duplication) then the accepted packets will only form a prefix of the heap and it'll be incomplete. Handling of incomplete heaps is subject to some other configuration options, but there is nothing specific to allow_out_of_order in it.

bmerry:
Thanks for the review. The suggestion of "getting started" documentation is a good one, but just requires the time to write it.

> In recv.py in katfgpu there is a AsyncRingbuffer class.

That's somewhat specific to katfgpu. katfgpu is a special case because it's a Python extension that uses the spead2 C++ API for improved performance. It's not the only one though (e.g. katsdpbfingest is similar) and I'm sure there will be more in future, and one day I need to streamline the process of writing such extensions (currently they involve a lot of cut-and-paste and usage of undocumented internals).

PR #95, opened 2020-09-28
Drop Python 3.5 support

bmerry:
It's about to be end-of-life, and removing it also allows dropping the explicit `loop` argument (passing it down to asyncio functions is deprecated in recent versions of Python). I figure this is a good time to remove that argument since I'm making other breaking changes in spead2 3.0.

It's probably worth reviewing the commits individually, because the pyupgrade commit is long but completely mechanical.

### DIFF
--- a/spead2/__init__.py
+++ b/spead2/__init__.py
@@ -227,7 +227,7 @@ class Descriptor:
                 fields.append('S1')
             else:
                 if code not in ['u', 'i', 'b']:
-                    raise ValueError('illegal format ({}, {})'.format(code, length))
+                    raise ValueError(f'illegal format ({code}, {length})')
                 fields.append('O')
         return _np.dtype(','.join(fields))
 
@@ -460,12 +460,12 @@ class Item(Descriptor):
                 if code == 'u':
                     raw = int(field)
                     if raw < 0 or raw >= (1 << length):
-                        raise ValueError('{} is out of range for u{}'.format(raw, length))
+                        raise ValueError(f'{raw} is out of range for u{length}')
                 elif code == 'i':
                     top_bit = 1 << (length - 1)
                     raw = int(field)
                     if raw < -top_bit or raw >= top_bit:
-                        raise ValueError('{} is out of range for i{}'.format(field, length))
+                        raise ValueError(f'{field} is out of range for i{length}')
                     # convert to 2's complement
                     if raw < 0:
                         raw += 2 * top_bit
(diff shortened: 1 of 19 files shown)


spead2/__init__.py
```diff
@@ -227,7 +227,7 @@ def _parse_format(cls, fmt):
                 fields.append('S1')
             else:
                 if code not in ['u', 'i', 'b']:
-                    raise ValueError('illegal format ({}, {})'.format(code, length))
+                    raise ValueError(f'illegal format ({code}, {length})')
```
ludwigschwardt, line 230: Nice to see automatic upgrades to f-strings. I should run pyupgrade on katpoint too...

spead2/test/test_recv.py
```diff
@@ -539,7 +538,7 @@ def test_no_stop_on_stop_item(self):
             [
                 self.flavour.make_plain_descriptor(
                     0x1234, 'test_string', 'a byte string', [('c', 8)], [None]),
-                Item(0x1234, 'Hello world'.encode('ascii'))
```
ludwigschwardt, line 542: Interesting that `.encode('ascii')` was needed in the first place.

spead2/test/test_passthrough_asyncio.py
```diff
@@ -30,10 +30,12 @@
...
-        ret = self.loop.run_until_complete(
-            self.transmit_item_groups_async(item_groups, memcpy, allocator, new_order))
-        self.loop.close()
+        loop = asyncio.new_event_loop()
+        try:
+            return loop.run_until_complete(
+                self.transmit_item_groups_async(item_groups, memcpy, allocator, new_order))
+        finally:
+            loop.close()
         return ret
```
ludwigschwardt, line 39: There is no `ret` to return. You can probably just remove the statement, or is it an error?

ludwigschwardt (changes requested):
Just one niggle (although I haven't searched for any errant loop variables in the rest of the codebase).

spead2/test/test_recv.py
bmerry: I think it might date from Python 2 support - possibly before `b'Hello world'` was accepted in Python 2.

spead2/test/test_passthrough_asyncio.py
bmerry: I've already fixed this in a subsequent PR. I'd rather not deal with trying to backport it.

ludwigschwardt: Saw that in passing 👍

ludwigschwardt (approved):

bmerry:
Merged manually (by merging a downstream branch), so closing.

PR #96, opened 2020-09-28
Convert unit tests to pytest

bmerry:
This is a relatively lightweight conversion, mainly to move from asynctest (which doesn't seem to be maintained) to pytest-asyncio. Unittest assertions are replaced by assertion statements, but the class structure is still in place even though a lot of it would be more gracefully handled by parametrisation and fixtures.

The directory structure is also not changed, and that causes some problems with pytest testing the local source tree rather than the installed version. I'll fix that in a follow-up PR so that I'm not mixing code changes with renames.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -165,11 +165,11 @@ setup(
     ],
     tests_require=[
         'netifaces',
-        'nose',
-        'asynctest'
+        'pytest',
+        'pytest-asyncio',
+        'pytest-timeout'
     ],
     python_requires='>=3.6',
-    test_suite='nose.collector',
     packages=find_packages(),
     package_data={'': ['py.typed', '*.pyi']},
     entry_points={
--- a/spead2/test/test_passthrough.py
+++ b/spead2/test/test_passthrough.py
@@ -21,8 +21,7 @@ import sys
 
 import numpy as np
 import netifaces
-from nose.tools import assert_equal, assert_raises, timed
-from nose.plugins.skip import SkipTest
+import pytest
 
 import spead2
 import spead2.send
@@ -30,17 +29,17 @@ import spead2.recv
 
 
 def _assert_items_equal(item1, item2):
-    assert_equal(item1.id, item2.id)
-    assert_equal(item1.name, item2.name)
-    assert_equal(item1.description, item2.description)
-    assert_equal(item1.shape, item2.shape)
-    assert_equal(item1.format, item2.format)
+    assert item1.id == item2.id
+    assert item1.name == item2.name
+    assert item1.description == item2.description
+    assert item1.shape == item2.shape
+    assert item1.format == item2.format
     # Byte order need not match, provided that values are received correctly
     if item1.dtype is not None and item2.dtype is not None:
-        assert_equal(item1.dtype.newbyteorder('<'), item2.dtype.newbyteorder('<'))
+        assert item1.dtype.newbyteorder('<') == item2.dtype.newbyteorder('<')
     else:
-        assert_equal(item1.dtype, item2.dtype)
-    assert_equal(item1.order, item2.order)
+        assert item1.dtype == item2.dtype
+    assert item1.order == item2.order
     # Comparing arrays has many issues. Convert them to lists where appropriate
     value1 = item1.value
     value2 = item2.value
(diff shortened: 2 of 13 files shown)


spead2/test/test_recv.py
```diff
@@ -411,16 +409,16 @@ def test_duplicate_item_pointers(self):
...
+        assert len(items) == 2
         items.sort(key=lambda item: item.id)
-        assert_equal(0x1600, items[0].id)
-        assert_true(items[0].is_immediate)
-        assert_equal(12345, items[0].immediate_value)
-        assert_equal(0x5000, items[1].id)
-        assert_false(items[1].is_immediate)
-        assert_equal(payload, bytearray(items[1]))
+        assert items[0].id == 0x1600
+        assert items[0].is_immediate is True
```
ludwigschwardt, line 417: Why `is True`? Can it be non-boolean? I guess the strictness gives a bit of extra protection.

spead2/test/test_passthrough.py
```diff
@@ -21,51 +21,42 @@
...
-def timed_class(cls):
-    """Class decorator version of `nose.tools.timed`"""
-    for key in cls.__dict__:
-        if key.startswith('test_'):
-            setattr(cls, key, timed(2)(getattr(cls, key)))
-    return cls
-
-
-@timed_class
+@pytest.mark.timeout(5)
```
ludwigschwardt, line 59: Nice.

spead2/test/test_recv_asyncio.py
```diff
@@ -13,19 +13,19 @@
...
 
 
-class TestRecvAsyncio(asynctest.TestCase):
+@pytest.mark.asyncio
+class TestRecvAsyncio:
     async def test_async_iter(self):
         tp = spead2.ThreadPool()
         queue = spead2.InprocQueue()
-        sender = spead2.send.InprocStream(tp, queue)
+        sender = spead2.send.InprocStream(tp, [queue])
```
ludwigschwardt, line 28: Did this interface change too?

spead2/test/test_send_asyncio.py
```diff
@@ -16,66 +16,58 @@
...
         for i in range(5):
             asyncio.ensure_future(self.stream.async_send_heap(self.heap))
-        # The above only queues up the async sends on the event loop. The rest of the
-        # test needs to be run from inside the event loop
-        asyncio.get_event_loop().run_until_complete(self._test_async_flush())
-
-    async def _test_send_error(self, future):
-        with assert_raises(IOError):
-            await future
+        await self._test_async_flush()
```
ludwigschwardt, line 55: Looks cleaner than the old asynctest 🎉

setup.py
```diff
@@ -165,11 +165,11 @@ def build_extension(self, ext):
...
     tests_require=[
         'netifaces',
-        'nose',
-        'asynctest'
+        'pytest',
+        'pytest-asyncio',
+        'pytest-timeout'
     ],
     python_requires='>=3.6',
-    test_suite='nose.collector',
```
ludwigschwardt, line 172: Unrelated question: so `pytest` doesn't need a `test_suite` entry?

ludwigschwardt (approved):
Good stuff!

spead2/test/test_recv.py
bmerry: Yeah, just ensures that it's really boolean rather than a truthy non-boolean.

setup.py
bmerry: It's more that 'setup.py test' is deprecated (e.g. see https://github.com/pytest-dev/pytest/issues/5534).

spead2/test/test_recv_asyncio.py
bmerry: The old code is a deprecated interface, and pytest helpfully reports DeprecationWarnings which is how I spotted it.

setup.py
ludwigschwardt: Ah, right. I've just stumbled on that in NumPy's `setup.py` as I was trying to reach enlightenment 🙂

spead2/test/test_recv_asyncio.py
ludwigschwardt: I'll keep that in mind for katpoint too, thanks for the tips.

bmerry:
Merged manually (by merging a downstream branch), so closing.

PR #102, opened 2020-10-13
Use IP checksum offloading where available 

bmerry:
It makes about a 5% performance improvement when using 1024B packets.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -13,6 +13,8 @@ Other changes:
 - Allow ibverbs senders to register memory regions for zero-copy
   transmission.
 - Add C++ preprocessor defines for the version number.
+- Use IP/UDP checksum offloading for sending with ibverbs (improves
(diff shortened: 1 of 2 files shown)


doc/changelog.rst
```diff
@@ -13,6 +13,8 @@ Other changes:
 - Allow ibverbs senders to register memory regions for zero-copy
   transmission.
 - Add C++ preprocessor defines for the version number.
+- Use IP/UDP checksum offloading for sending with ibverbs (improves
```
gcallanan, line 16: From the changes I can see, only the IBV_SEND_IP_CSUM flag seems to be set. I see nothing indicating UDP checksum is being offloaded. How is the UDP offload functionality enabled?

bmerry: From the man page:

>       IBV_SEND_IP_CSUM  Offload the IPv4 and TCP/UDP checksum calculation.

gcallanan (approved):

PR #104, opened 2020-10-13
Overhaul command-line handling in C++ tools

bmerry:
Each of the C++ tools (spead2_recv, spead2_send, spead2_bench) had its own command-line handling, which lead to a lot of duplication (spead2_recv and spead2_bench had options controlling receivers, spead2_send and spead2_bench had options controlling senders), and there were inconsistencies (e.g. spead2_bench was missing the --ttl option).

There is now a helper file (plus header) where command-line options are registered and which has the logic to build streams from them. Somehow this has lead to an increase in the total lines of code, but it's also brought along some capabilities that were missing, and it'll require less work to add new options in future.

This just covers the C++ tools. The Python equivalents have the same issue, but I'll leave that for a later PR. There are also some misc unrelated fixups that I threw in as I went.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 global-include Makefile.in
 global-include aclocal.m4
-exclude src/common_features.h
+exclude include/spead2/common_features.h
 exclude python-build.cfg
 exclude .gitignore
 include configure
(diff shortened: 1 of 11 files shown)


bmerry:
I'd suggest starting with spead2_cmdline.h/.cpp. Note that a lot of the code in them has been moved from other places rather than being new code.

sratcliffe (approved):
Looks good to me. Probably not in a position to really scrutinise the approach, but structurally it looks good, the copied code looks accurate and no typos or other things I can find.

PR #109, opened 2020-10-16
Unify Python command-line argument parsing

bmerry:
This is like the previous PR, but for Python.

### DIFF
--- a/src/spead2/tools/bench_asyncio.py
+++ b/src/spead2/tools/bench_asyncio.py
@@ -313,6 +313,8 @@ def main():
     args = parser.parse_args()
     logging.basicConfig(level=getattr(logging, args.log.upper()))
     if 'endpoint' in args:
+        if args.send_ibv and not args.multicast:
+            parser.error('--send-ibv requires --multicast')
         receiver.mem_pool = True
         receiver.mem_lower = args.heap_size
         receiver.mem_upper = args.heap_size + 1024  # more than enough for overheads
--- a/src/spead2/tools/bench_asyncio.py
+++ b/src/spead2/tools/bench_asyncio.py
@@ -81,14 +81,8 @@ class AgentConnection:
                         setattr(protocol, key, value)
                     for (key, value) in command['receiver'].items():
                         setattr(receiver, key, value)
-                    if args.recv_affinity is not None and len(args.recv_affinity) > 0:
-                        spead2.ThreadPool.set_affinity(args.recv_affinity[0])
-                        thread_pool = spead2.ThreadPool(
-                            1, args.recv_affinity[1:] + args.recv_affinity[:1])
-                    else:
-                        thread_pool = spead2.ThreadPool()
                     stream = spead2.recv.asyncio.Stream(
-                        thread_pool,
+                        receiver.make_thread_pool(),
                         receiver.make_stream_config(),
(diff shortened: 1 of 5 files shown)


src/spead2/tools/cmdline.py
```diff
@@ -0,0 +1,321 @@
...
+        if _HAVE_IBV:
+            if self.ibv and not self.bind:
+                parser.error('--ibv requires --bind')
+            if self._protocol.tcp and self.ibv:
+                parser.error('--ibv and --tcp are incompatible')
+
+        if self.buffer is None:
+            if self._protocol.tcp:
+                self.buffer = spead2.recv.asyncio.Stream.DEFAULT_TCP_BUFFER_SIZE
+            elif _HAVE_IBV and self.ibv:
```
sratcliffe, line 167: I don't think self.ibv can be True unless _HAVE_IBV is already True, so looks like a redundant check (here and in the next stanza).

src/spead2/tools/bench_asyncio.py
```diff
@@ -281,72 +267,62 @@ def main():
...
-                        help='Maximum packet size to use for UDP [%(default)s]')
     master.add_argument('--heap-size', metavar='BYTES', type=int, default=4194304,
                         help='Payload size for heap [%(default)s]')
-    master.add_argument('--addr-bits', metavar='BITS', type=int, default=40,
-                        help='Heap address bits [%(default)s]')
     master.add_argument('--multicast', metavar='ADDRESS', type=str,
                         help='Send via multicast group [unicast]')
+    protocol.add_arguments(master)
     group = master.add_argument_group('sender options')
+    sender.add_arguments(group)
```
sratcliffe, line 302: [Comment for next line] - it looks to me like all examples of senders and receivers in the examples have an `affinity` flag (in the bench case explicitly predicated with send- and recv-). Couldn't these options be promoted to the common sender and receiver argument lists in `cmdline.py` ?

src/spead2/tools/cmdline.py
bmerry: The catch is that if `_HAVE_IBV` is false then `self.ibv` won't exist at all and you'll get an AttributeError.

src/spead2/tools/bench_asyncio.py
bmerry: Yes, I'll do that. I started just mirroring what was done in the C++ code, which for whatever reason doesn't have those flags (probably because they're single-threaded and hence numactl is adequate for controlling affinity).

src/spead2/tools/cmdline.py
sratcliffe: Sure - I didn't see a reason to have the init of self.ibv protected in the first place, but no doubt it breaks something somewhere else :)

bmerry: Yes, when it uses reflection in `_extract_args`, it'll fail to determine the value of the command-line argument because there is no such argument.

src/spead2/tools/bench_asyncio.py
bmerry: I've added a new commit that cleans that up.

sratcliffe (approved):
Nice - those affinity changes look good.

PR #112, opened 2020-10-19
Introduce recv.UdpIbvConfig

bmerry:
To mirror the send.UdpIbvConfig class, introduce one for the receive path. The previous API is marked deprecated but retained. Since there was a lot of commonality between the two UdpIbvConfig classes, a base class is introduced in common_ibv.cpp.

There are some other odds and ends mixed in here:
- Remove the _PybindStr alias from the .pyi files; it was really there to support Python 2.
- Bring back UdpIbvStream.DEFAULT_* constants as (deprecated) aliases for the constants in send.UdpIbvConfig.


include/spead2/common_ibv.h
```diff
@@ -405,6 +405,145 @@ class ibv_rwq_ind_table_t : public std::unique_ptr<ibv_rwq_ind_table, detail::ib
...
+public:
+    /// Get the configured endpoints
+    const std::vector<boost::asio::ip::udp::endpoint> &get_endpoints() const { return endpoints; }
+    /**
+     * Set the endpoints (replacing any previous).
+     *
+     * @throws std::invalid_argument if any element of @a endpoints is invalid.
+     */
+    Derived &set_endpoints(const std::vector<boost::asio::ip::udp::endpoint> &endpoints);
+    /**
```
gcallanan, line 435: Should there be a new line between `Derived &set_endpoints(const std::vector<boost::asio::ip::udp::endpoint> &endpoints);` and the block comment that follows? Or are you grouping `set_endpoint() and add_endpoint()` together?

bmerry: I'm grouping related functions together.

include/spead2/common_ibv.h
```diff
@@ -405,6 +405,145 @@ class ibv_rwq_ind_table_t : public std::unique_ptr<ibv_rwq_ind_table, detail::ib
 /// Construct a table with a single entry
 ibv_rwq_ind_table_t create_rwq_ind_table(const rdma_cm_id_t &cm_id, const ibv_wq_t &wq);
 
+namespace detail
+{
+
+/**
+ * Base for @ref spead2::recv::udp_ibv_config and @ref spead2::send::udp_ibv_config.
+ * It uses the curiously recursive template pattern to avoid being
+ * polymorphic (and to allow the return type of setters to be the derived class).
```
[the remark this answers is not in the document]
bmerry: I've updated the comment.

gcallanan (approved):

PR #120, opened 2020-10-20
Avoid running arena initialisation on the hot path

bmerry:
Profiling revealed that glibc initialises an arena the first time
dynamic allocation is one in a thread, and that this was happening when
the first packet arrived. To avoid that, do a dummy initialisation as
soon as the worker thread starts.

### DIFF
--- a/src/common_thread_pool.cpp
+++ b/src/common_thread_pool.cpp
@@ -37,6 +37,13 @@ static void run_io_service(boost::asio::io_service &io_service)
 {
     try
     {
+        /* Glibc's memory allocator seems to do some initialisation the
+         * first time a thread does a dynamic allocation. To avoid that
+         * occurring in a real-time situation, do a dummy allocation now to
+         * initialise it.
+         */
+        int *dummy = new int;
+        delete dummy;
         io_service.run();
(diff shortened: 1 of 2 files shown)


src/common_thread_pool.cpp
```diff
@@ -37,6 +37,13 @@ static void run_io_service(boost::asio::io_service &io_service)
 {
     try
     {
+        /* Glibc's memory allocator seems to do some initialisation the
+         * first time a thread does a dynamic allocation. To avoid that
+         * occurring in a real-time situation, do a dummy allocation now to
+         * initialise it.
+         */
+        int *dummy = new int;
+        delete dummy;
```
gcallanan, line 46: Does this have a noticeable impact on performance? Perhaps it reduces those initial few packet drops that were observed?

bmerry: I suspect it probably makes very little difference - early drops are more likely due to power management and/or the switch sorting out its multicast routing. I only noticed it because I decided to check where page faults were occurring and this was responsible for one page fault.

gcallanan (approved):

PR #128, opened 2020-10-29
Support interleaving packets from multiple heaps

bmerry:


bmerry:
@gcallanan I've done some cleanup on the commit history, so that reading the individual commits will make sense. I'd start with the end-user documentation commit (one of the last ones) first to understand the interface to the feature.

include/spead2/send_stream.h
```diff
@@ -199,7 +224,133 @@ class stream
...
+        for (Iterator it = first; it != last; ++it)
+        {
+            const heap &h = get_heap(*it);
+            s_item_pointer_t cnt = get_heap_cnt(*it);
+            std::size_t substream_index = get_heap_substream_index(*it);
+            if (substream_index >= num_substreams)
+            {
+                unwind.abort();
+                lock.unlock();
+                log_warning("async_send_heap: dropping heap because substream index is out of range");
```
gcallanan, line 289: Must this not now read "async_send_heaps" instead of "async_send_heap"?

This applies to a number of log_warning() calls in this file.

src/send_stream.cpp
```diff
@@ -141,53 +104,9 @@ bool stream::async_send_heap(const heap &h, completion_handler handler,
                              s_item_pointer_t cnt,
                              std::size_t substream_index)
```
gcallanan, line 105: So async_send_heap() now just wraps a call to async_send_heaps() essentially?

src/py_send.cpp
```diff
@@ -146,12 +146,32 @@ class stream_wrapper : public Base
         else
             return state->bytes_transferred;
     }
+
+    /// Sends multiple heaps synchronously
+    item_pointer_t send_heaps(const std::vector<heap_reference> &heaps, group_mode mode)
+    {
+        // See comments in send_heap
+        auto state = std::make_shared<callback_state>();
+        Base::async_send_heaps(
```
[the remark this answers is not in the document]
bmerry: The Python API has been a synchronous and an asynchronous interface. You probably aren't going to get top performance out of the synchronous interface, but it's much easier to use. For example, I think the CMC uses it. This is just extending the synchronous send_heap function to the plural.

include/spead2/send_stream.h
```diff
@@ -43,98 +48,116 @@ namespace spead2
...
-    /// Get rate-limiting method
-    rate_method get_rate_method() const { return method; }
-
-    /// Get product of rate and burst_rate_ratio
-    double get_burst_rate() const;
-
-    stream_config();
+    const send::heap &heap;
+    s_item_pointer_t cnt;
+    std::size_t substream_index;
```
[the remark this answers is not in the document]
bmerry: The heap counter (SPEAD item 0x0001).

include/spead2/send_stream.h
bmerry: This code is shared between the two. I can change it to `async_send_heap(s)`?

src/send_stream.cpp
bmerry: Essentially yes, except that because it uses `null_unwinder` instead of `unwinder` it gets to bypass some bits of code that are only needed when dealing with more than one heap.

include/spead2/send_stream.h
gcallanan: async_send_heap(s) may lead to less confusion, but that may just be personal preference. So whichever you prefer.

bmerry: I agree, async_send_heap(s) is better - I've implemented it. I also thought about trying to pass in the appropriate name as a parameter to the function, but this is a very performance-critical function so I didn't want to add any overheads.

gcallanan (approved):

PR #130, opened 2020-11-10
Improve send performance by eliminating a memory allocation

bmerry:
Instead of packet_generator::next_packet allocate memory to store data that it can't reference in place (the SPEAD header and item pointers), require the caller to provide memory for this. The caller can allocate memory for each packet in flight and recycle it.

A bonus benefit is that for the ibv sender, the packet generator will write the SPEAD header and item pointers directly into the payload area where it is needed, avoiding the need to copy it.

This improves send performance with 1KB packets roughly from 64 -> 86 Gb/s.

### DIFF
--- a/include/spead2/send_packet.h
+++ b/include/spead2/send_packet.h
@@ -81,8 +81,20 @@ private:
 public:
     packet_generator(const heap &h, item_pointer_t cnt, std::size_t max_packet_size);
 
+    std::size_t get_max_packet_size() const;
+
     bool has_next_packet() const;
-    packet next_packet();
+    /**
+     * Create a packet ready for sending on the network. The caller must
+     * provide space for storing data, of size at least the @a max_packet_size
+     * passed to the constructor, and with at least item_pointer_t alignment.
+     * The returned buffer sequence may contain a mix of pointers to the
+     * original heap and pointers within the scratch area.
+     *
+     * If there are no more packets to send for the heap, an empty list is
+     * returned.
+     */
+    std::vector<boost::asio::const_buffer> next_packet(std::uint8_t *scratch);
 };
 
 } // namespace send
--- a/include/spead2/send_packet.h
+++ b/include/spead2/send_packet.h
@@ -35,24 +35,6 @@ namespace send
 
 class heap;
 
-/**
- * A packet ready for sending on the network. It contains some internally
- * held data, and const buffer sequence that contains a mix of pointers to
- * the internal data and pointers to the heap's items.
- *
- * If @a buffers is empty, it indicates the end of the heap.
- *
- * @todo Investigate whether number of new calls could be reduced by using
- * a pool for the case of packets with no item pointers other than the
- * per-packet ones; or by having the caller of the packet_generator provide
- * storage.
- */
-struct packet
-{
-    std::unique_ptr<std::uint8_t[]> data;
-    std::vector<boost::asio::const_buffer> buffers;
-};
-
 class packet_generator
 {
 private:
--- a/include/spead2/send_packet.h
+++ b/include/spead2/send_packet.h
@@ -42,36 +42,49 @@ private:
     static constexpr std::size_t prefix_size = 8 + 4 * sizeof(item_pointer_t);
 
     const heap &h;
-    item_pointer_t cnt;
-    std::size_t max_packet_size;
-    std::size_t max_item_pointers_per_packet;
+    const item_pointer_t cnt;
+    const std::size_t max_packet_size;
+    const std::size_t max_item_pointers_per_packet;
 
+    /// @name Item pointer generation
+    /// @{
     /// Next item pointer to send
     std::size_t next_item_pointer = 0;
+    /// Address at which payload for the next item will be found
+    std::size_t next_address = 0;
+    /// @}
+
+    /// @name Payload generation
+    /// @{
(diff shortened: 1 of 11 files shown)


src/send_udp.cpp
```diff
@@ -43,9 +43,15 @@ class udp_writer : public writer
 #if SPEAD2_USE_SENDMMSG
     struct mmsghdr msgvec[max_batch];
     std::vector<struct iovec> msg_iov;
-    transmit_packet packets[max_batch];
+    struct
+    {
+        transmit_packet packet;
+        std::unique_ptr<std::uint8_t[]> scratch;
+    } packets[max_batch];
```
gcallanan, line 50: It almost seems like in this case scratch should be a parameter within the packet object. I imagine its kept separate because transmit packet is really only applicable when SPEAD2_USE_SENDMMSG is defined?

src/send_tcp.cpp
```diff
@@ -159,7 +161,8 @@ tcp_writer::tcp_writer(
     socket(make_socket(get_io_service(), endpoints, buffer_size, interface_address)),
     pre_connected(false),
     endpoint(endpoints[0]),
-    connect_handler(std::move(connect_handler))
+    connect_handler(std::move(connect_handler)),
+    scratch(new std::uint8_t[config.get_max_packet_size()])
```
gcallanan, line 165: So if I understand this correctly, enough scratch space is allocated for a single header? So basically only a single packet can be under construction or being transmitted in the tcp case? I cant quite recall how the TCP case works. Can multiple packets be queued up at any one time?

src/send_inproc.cpp
```diff
@@ -32,21 +32,22 @@ namespace send
...
-    boost::asio::buffer_copy(buffer, in.buffers);
+    boost::asio::buffer_copy(buffer, in);
     return out;
 }
 
 class inproc_writer : public writer
 {
 private:
     std::vector<std::shared_ptr<inproc_queue>> queues;
+    std::unique_ptr<std::uint8_t[]> scratch;   ///< Scratch space for constructing packets
```
gcallanan, line 50: Maybe this should read: "For storing the packet header"? Constructing packets seems to imply the whole packet goes here. This would apply to all the different writes floating around with this comment.

gcallanan (commented):
I have left a few comments. 

I am a bit confused as to the function of the scratch patch. I have internally assigned it the roll of storing packet header information, but that only makes sense in the ibverbs case. Does the scratch patch just store a whole packet in most cases?

When I am clear on this,  I will look at it again so see if it makes sense to me.

bmerry:
> I am a bit confused as to the function of the scratch patch. I have internally assigned it the roll of storing packet header information, but that only makes sense in the ibverbs case. Does the scratch patch just store a whole packet in most cases?

It's for anything that the packet generator needs to synthesize rather than being able to point at in existing memory. There are three things that can go in there:
- the SPEAD header (8 bytes)
- the item pointers
- dummy payload, if there isn't any real payload and we have to make some up so that each packet has a different payload offset (rare).
I'm not sure how best to summarise that in a few words for comments.

Having the packet generator return a gather list rather than putting all the packet data into contiguous memory is useful beyond just the ibverbs and sendmmsg cases. For standard boost::asio socket networking one can pass a gather list, which underneath will call sendmsg and the gathering will be done as part of the user→kernel copy.

include/spead2/send_packet.h
```diff
@@ -35,54 +35,61 @@ namespace send
...
- * per-packet ones; or by having the caller of the packet_generator provide
- * storage.
- */
-struct packet
-{
-    std::unique_ptr<std::uint8_t[]> data;
-    std::vector<boost::asio::const_buffer> buffers;
-};
-
 class packet_generator
```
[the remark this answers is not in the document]
bmerry: > Should there be comment describing what this class does?

Probably, but I think beyond the scope of this PR.

include/spead2/send_packet.h
```diff
@@ -35,54 +35,61 @@ namespace send
...
+    std::size_t get_max_packet_size() const;
+
     bool has_next_packet() const;
-    packet next_packet();
+    /**
+     * Create a packet ready for sending on the network. The caller must
+     * provide space for storing data, of size at least
+     * @ref get_max_packet_size, and with at least @c item_pointer_t alignment.
+     * The first element of the returned buffer sequence will reference a
+     * prefix of @a scratch, while the remaining elements will reference to the
```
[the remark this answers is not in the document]
bmerry: I mean prefix in the sense of a contiguous range that starts with the first byte.

L2-4 headers aren't part of `scratch` - they're outside the scope of the packet generator, which just deals with L4 content.

src/send_tcp.cpp
bmerry: The TCP code currently only works on one packet at a time, which is why there only needs to be one scratch area for the whole class, whereas the UDP sender with sendmmsg has multiple.

src/send_udp.cpp
bmerry: Keeping it separate allows each writer class to decide on the memory management e.g. here there is a separate call to `new` for each packet, but the ibv sender just has a single verbs memory region and chops it up into space for each packet.

gcallanan (approved):

PR #132, opened 2020-11-17
Add a code example that assembles heaps directly to GPU memory

bmerry:
This serves mainly as an example for custom memory allocators and custom copy functions.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -158,7 +158,7 @@ SPEAD2_ARG_WITH(
 
 SPEAD2_ARG_WITH(
     [posix-semaphores],
-    [AS_HELP_STRING([--without-posix-semaphores], [Do not POSIX semaphores, even if available])],
+    [AS_HELP_STRING([--without-posix-semaphores], [Do not use POSIX semaphores, even if available])],
     [SPEAD2_USE_POSIX_SEMAPHORES],
     [SPEAD2_CHECK_FEATURE(
         [posix_semaphores], [POSIX semaphores],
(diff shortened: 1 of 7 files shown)


gcallanan:
So this basically has the potential to eliminates a memory copy from the receive side? A packet is copied into main memory and then copied to the heap on the GPU instead of system memory.

Does this work on the GTX/RTX cards? I remember testing gdrcopy out - it did not run on dbelab07(Maybe I was doing something wrong on my side).  Does it work straight out the box? Or is there some config that needs to be done first?

bmerry:
> So this basically has the potential to eliminates a memory copy from the receive side? A packet is copied into main memory and then copied to the heap on the GPU instead of system memory.

Exactly.

> Does this work on the GTX/RTX cards?

I don't know about the RTX 30xx (maybe the new RTX IO feature means it will work), but I've not managed to get it to work on the older cards, and people who've asked in issues have been told that it's not supported.

> Does it work straight out the box? Or is there some config that needs to be done first?

There is a kernel module (included with gdrcopy) that needs to be built and loaded.

gcallanan:
> I don't know about the RTX 30xx (maybe the new RTX IO feature means it will work), but I've not managed to get it to work on the older cards, and people who've asked in issues have been told that it's not supported.

Which GPUs have you tested it on?

bmerry:
> Which GPUs have you tested it on?

GTX 1080 Ti recently; I've probably tried on some older GPUs in the past.

gcallanan (approved):
This is a very interesting addition to the library. I hope that one day we are able to use it. 

This example program will be a useful starting point in the future.

PR #133, opened 2020-11-18
Add an example of GPUDirect RDMA for sending 

bmerry:
Data is sent directly from the GPU to the network.


gcallanan (approved):
This implementation is far simpler than the receiver implementation. No need for a gdr_mh_t struct and the associated mapping functions.

PR #134, opened 2020-11-20
Fix a memory leak when receiving with ibverbs

bmerry:
The destruction order was wrong. Trying to destroy ibverbs objects that
are still referenced by other objects fails with EBUSY. The completion
queue buffer can be quite substantial (e.g. 128MB).


sratcliffe:
I know you didn't ask me to review, but the changes look sensible to me.

gcallanan (approved):

bmerry:
Thanks to both of you.

PR #135, opened 2020-11-24
Always SFENCE at end of memcpy_nontemporal

bmerry:
It might be needed for correct ordering with write-combining memory,
such as used by gdrcopy.

### DIFF
--- a/src/common_memcpy.cpp
+++ b/src/common_memcpy.cpp
@@ -43,7 +43,14 @@ void *memcpy_nontemporal(void * __restrict__ dest, const void * __restrict__ src
     {
         if (head >= n)
         {
-            return std::memcpy(dest_c, src_c, n);
+            std::memcpy(dest_c, src_c, n);
+            /* Not normally required, but if the destination is
+             * write-combining memory then this will flush the combining
+             * buffers. That may be necessary if the memory is actually on
+             * a GPU or other accelerator.
+             */
+            _mm_sfence();
+            return dest;
         }
         std::memcpy(dest_c, src_c, head);
         dest_c += head;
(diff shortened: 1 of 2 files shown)


src/common_memcpy.cpp
```diff
@@ -43,7 +43,14 @@ void *memcpy_nontemporal(void * __restrict__ dest, const void * __restrict__ src
...
         if (head >= n)
         {
-            return std::memcpy(dest_c, src_c, n);
+            std::memcpy(dest_c, src_c, n);
+            /* Not normally required, but if the destination is
+             * write-combining memory then this will flush the combining
+             * buffers. That may be necessary if the memory is actually on
+             * a GPU or other accelerator.
+             */
+            _mm_sfence();
```
gcallanan, line 52: Does _mm_sfence affect performance normally?

Where do these "combining buffers" live? Is this in the CPU cache?

bmerry: It'll affect performance very slightly because it prevents the CPU doing certain things out-of-order. I don't expect this line of code to execute typically though, because it's only reached if the entire copy fits in one cache-line.

Combining buffers are part of the CPU but separate from the cache. See section 7.5.2 in https://www.amd.com/system/files/TechDocs/24593.pdf. I'm not sure I fully understand it, but I think WC buffers get used if the streaming instructions are used (which is what most of this function uses, but not this codepath) OR if the page tables mark the pages as write-combining.

gcallanan: What made you think to put this in? Was there any specific issue that was occurring or was it recommended in some manual?

bmerry: I haven't seen any issues, but [this](https://github.com/NVIDIA/gdrcopy/blob/ba82488d137454228faf32ef593d0ed912c486be/src/memcpy_avx.c#L110-L112) is probably what made me think of doing it.

gcallanan (approved):

PR #140, opened 2021-06-21
Use 'python3' everywhere to invoke Python

bmerry:
Out of the box, Ubuntu 20.04 has a /usr/bin/python3 but no
/usr/bin/python (and on older distributions, /usr/bin/python in Python
2). While virtual environments are preferred, using python3 for scripts
makes it possible to bootstrap and compile spead2 for C++ without
knowing much of anything about Python.


james-smith-za (approved):
Seems rational. Will probably increase the robustness, to be explicit about wanting python3.

PR #141, opened 2021-06-24
Add pipe option to CXXFLAGS

james-smith-za:
This makes compilation somewhat more speedious because you can use make
with the -j flag more effectively on multicore servers.

I am not familiar with the autoconf way of doing things so I may have in my
naivete done this wrong somehow. But it seems to work.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -30,7 +30,7 @@ AC_LANG(C++)
 ### Check for -std=c++11
 
 save_CXXFLAGS="$CXXFLAGS"
-CXXFLAGS="-std=c++11 $CXXFLAGS"
+CXXFLAGS="-std=c++11 $CXXFLAGS -pipe"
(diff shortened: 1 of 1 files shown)


bmerry:
This looks like the wrong place to me. It's just a check that passing the `-std=c++11` flag to the compiler (and trying to compile code using a c++11 feature) succeeds, after which CXXFLAGS is restored a few lines lower down. I think you need to add it to AM_CXXFLAGS in `src/Makefile.inc.am`.

james-smith-za:
You are correct, I had it in the wrong place. If I add it where you suggest, the same options seem to get passed to nvcc which doesn't know what the `-pipe` option means.

I'm not really in the mood to try and disentangle the options passed to nvcc from those passed to gcc, so I am going to leave it as-is.

Watching htop while running `make -j` lets me see that all cores aren't fully engaged, but simply adding the `-j` flag gives a speedup of about 16x even without the `-pipe` option (2m18s vs 16s). So that's probably good enough.

PR #142, opened 2021-07-02
Fix some compiler warnings about unused parameters

bmerry:
Stub implementations of some abstract functions didn't use the
parameters, causing compiler warnings in some cases. Fix by deleting the
names of the parameters.

### DIFF
--- a/include/spead2/send_stream.h
+++ b/include/spead2/send_stream.h
@@ -80,12 +80,12 @@ static inline const heap &get_heap(const heap &h)
     return h;
 }
 
-static inline s_item_pointer_t get_heap_cnt(const heap &h)
+static inline s_item_pointer_t get_heap_cnt(const heap &)
 {
     return -1;
 }
 
-static inline std::size_t get_heap_substream_index(const heap &h)
+static inline std::size_t get_heap_substream_index(const heap &)
 {
     return 0;
 }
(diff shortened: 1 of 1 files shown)


james-smith-za (commented):
Without having delved into the depths of `spead2`, I may be missing something.

But bare parameters without names make me nervous. If they're not used, why not just remove them completely? These don't seem to be derived types so they're not overriding some base method with a defined set of arguments.

Are they expected to be derived? There's nothing in the class declarations to make me think it's an abstract base class. If so, how is the derived class's author supposed to know what the parameter does?

bmerry:
> But bare parameters without names make me nervous. If they're not used, why not just remove them completely? These don't seem to be derived types so they're not overriding some base method with a defined set of arguments.

Function overloading: the same functions can be called with either a `heap` (these changes) or a `heap_reference` (lines 93-106) and the behaviour depends on the type of the parameter. There are functions elsewhere that can take either (using C++ templates) and use these functions to extract information.

james-smith-za:
Right. I didn't see that the functions were overloaded, it didn't show up in the diff. I think I get it.
So an equivalent kind of logic in Python would be something like this:
```python
def get_heap_cnt(h: Union[heap, heap_reference]):
    if isinstance(h, heap):
        return -1
    elif isinstance(h, heap_reference):
        return h.ref.cnt
``` 

So in the case of the anonymous function argument, only the type of the argument matters, to let the compiler know which overloaded function to use. The nature of the `heap` doesn't matter, just that it's a `heap` and not a `heap_reference`.

Did I understand correctly?

james-smith-za (approved):

bmerry:
> Did I understand correctly?

Yep.

PR #143, opened 2021-07-07
Some minor packaging updates

bmerry:
- Use [pypa-build](https://pypa-build.readthedocs.io/en/latest/) to generate release artefacts.
- Use setuptools_scm to ensure that the right files go into the release artefacts (previously I relied on `setuptools_git` in the host environment, but pypa-build performs build isolation so something needs to be listed in pyproject.toml). Note that setuptools_scm is used just for this functionality and not to determine the version.
- Bump pybind11 to the latest version


src/spead2/_version.py
```diff
@@ -1 +1 @@
-__version__ = "3.2.2"
+__version__ = "3.2.3"
```
james-smith-za, line 1: Should this file be in version control? Doesn't `setuptools_scm` generate it?
Or is this specifically for sdists?

james-smith-za (approved):
Looks good to me. One question mostly demonstrating my lack of familiarity with best practices for distribution of Python packages.

src/spead2/_version.py
bmerry: As noted in the introduction:
> Note that setuptools_scm is used just for this functionality and not to determine the version.

This is the primary source of truth of the version number for spead2, and tags are made after cutting a release.

PR #144, opened 2021-07-09
Add spead2_net_raw tool 

bmerry:
This simplifies the process of running programs with CAP_NET_RAW. See
the updates in doc/tools.rst for further explanation.


doc/changelog.rst
```diff
@@ -1,8 +1,9 @@
 Changelog
 =========
 
-.. rubric:: 3.2.3
+.. rubric:: 3.3.0
```
james-smith-za, line 4: So if I'm understanding correctly, the LSD is bugfixes or incremental changes, the second-to-last digit is incremented when new features are added, and the major version is when changes break backwards compatibility?

james-smith-za (approved):

doc/changelog.rst
bmerry: Yep. Good old semantic versioning. spead2_net_raw is a new feature, hence minor version is updated.

PR #148, opened 2021-07-14
Migrate from Travis CI to Github Actions

bmerry:
There are a few commits to clean up problems that only showed up with Github Actions.


bmerry:
@james-smith-za assigning to you since you expressed interest in Github Actions, but let me know if you'd rather I found someone else to review.

james-smith-za (approved):
Interesting, thanks - I will freely concede to not knowing except in a hand-wavy way what CI systems do, so I can't really comment either way.

I have a few observations - the way of supporting multiple Python versions was quite interesting, it looked as though in `.github/workflows/test.yml` that the default was Python 3.8, and that you have to add explicit tests to do 3.6, 7 and 9. If I understood correctly what was going on there.

I'm somewhat puzzled by the desire to support PyPy as well. I wasn't aware of any spead2 users who insisted that CPython isn't good enough.

bmerry:
> I have a few observations - the way of supporting multiple Python versions was quite interesting, it looked as though in .github/workflows/test.yml that the default was Python 3.8, and that you have to add explicit tests to do 3.6, 7 and 9. If I understood correctly what was going on there.

I think you have to specify *something* when using the `setup-python` action. I chose 3.6 and 3.9 as the oldest supported version and the newest version, on the assumption that if it works on the boundaries it'll probably work on anything in between.

> I'm somewhat puzzled by the desire to support PyPy as well. I wasn't aware of any spead2 users who insisted that CPython isn't good enough.

That's true. It come out of an experiment at a PyCon ZA sprint a few years ago, where I decided to see what would be needed to support PyPy, as a way of learning a bit more about it.

PR #149, opened 2021-07-15
Rename spead2_net_raw.c to spead2_net_raw.cpp

### LINKED ISSUE #146: Use C++ compiler flags for spead2_net_raw
While it's written in pure C, it would simplify matters to make spead2_net_raw a .cpp file so that it gets all the compiler flag setup that's done for C++ code.

bmerry:
Closes #146.

### DIFF
--- a/src/spead2_net_raw.cpp
+++ b/src/spead2_net_raw.cpp
@@ -14,6 +14,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+/* This file is written in C. It's in a .cpp file so that the build system
+ * doesn't need to worry about setting the right compiler flags for both C
+ * and C++.
+ */
+
 #include <stdio.h>
 #include <stddef.h>
 #include <stdlib.h>
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,6 @@ AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_SRCDIR([src/common_logging.cpp])
 AC_CONFIG_MACRO_DIRS([m4])
 AM_INIT_AUTOMAKE([foreign])
-AC_PROG_CC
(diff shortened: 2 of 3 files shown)


src/spead2_net_raw.cpp
```diff
@@ -14,6 +14,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+/* This file is written in C. It's in a .cpp file so that the build system
+ * doesn't need to worry about setting the right compiler flags for both C
+ * and C++.
+ */
+
```
james-smith-za, line 21: I may be betraying my ignorance here, but the way I understand this is as follows:
- C and C++ compilers end up making symbols differently, C++ mangles them so you have to put `extern C` if you are pulling in C headers (and libraries) to a C++ program.
- However, C++ is a superset of C, so (in most cases) you can just write C code and compile it with a C++ compiler and get a valid output. It will have different symbol names but that won't matter, since it'll have been compiled and linked as though it were C++.

If this is the last outstanding thing that was using the C compiler then it makes sense to do this, I think. Did I understand correctly?

james-smith-za (approved):

src/spead2_net_raw.cpp
bmerry: All pretty much true. C++ isn't *precisely* a superset of C (for one thing, C code that uses C++ keywords as identifiers won't compile, but I think there are also some corner cases depending on exactly which versions of C and C++ you're referring to), but for practical purposes it is.

spead2_net_raw.c was the first and only C file in spead2, so this takes spead2 back to not having any C.

james-smith-za: >for one thing, C code that uses C++ keywords as identifiers won't compile

Ah right, good point. I didn't think about that, but useful to bear in mind.

PR #153, opened 2021-08-26
Chunking receive

bmerry:
Provides a fast path for Python to handle high heap rates by collecting multiple heaps into a single larger buffer.


james-smith-za (changes requested):
Some small suggestions for correction / clarification for me.

Without being much more familiar with spead2's internals than I am (and without stronger c++-fu), I don't think I'll be able to make more of a contribution than this.

james-smith-za (approved):

bmerry:
Thanks for the review - I've taken all your suggestions.

PR #154, opened 2021-09-06
Use cibuildwheel to automate wheel-building

bmerry:
This makes it less likely to depend on what versions of tools I happen to have on my system at the time, and also means it gets tested on every commit. In addition to building wheels, I also build an sdist and extract the debug symbols.

I haven't automated pushing any of this to PyPI or Github's releases. I prefer to still have a human in the loop for that to sanity check things.


manylinux/before_all.sh
```diff
@@ -0,0 +1,36 @@
...
+
+package="$1"
+
+yum install -y \
+    wget libpcap libpcap-devel python-devel \
+    cmake3 ninja-build pandoc libnl3-devel \
+    ccache
+
+# Install boost
+wget --progress=dot:mega https://ufpr.dl.sourceforge.net/project/boost/boost/1.74.0/boost_1_74_0.tar.bz2 -O /tmp/boost_1_74_0.tar.bz2
```
ludwigschwardt, line 15: Why from source and not via `yum`?

ludwigschwardt (approved):

manylinux/before_all.sh
bmerry: The build is done in a Docker image that deliberately runs an old version of CentOS, so that you only end up depending on old symbols, but I wanted an up-to-date version of Boost.

PR #155, opened 2021-09-16
Custom statistics

bmerry:
There are basically three things happening here, but it unfortunately turns into a lot of boilerplate and a very large PR:

1. New user-defined can be registered on a stream (via the StreamConfig setup), and accessed in chunk streams via an extra field in chunk_place_data.
2. To reflect the fact that the set of statistics is now a runtime property, StreamStats exposes dict-like and sequence-like interfaces for looking up and iterating over statistics. To reduce the cost of always finding a statistic by name, each statistic has an associated "index" (the position in the list of all statistics for the stream) and this can be used for quick lookup. The original attribute access is retained for backwards compatibility.
3. Chunk streams provide two new statistics, too_old_heaps and rejected_heaps.

In terms of the implementation, the `stream` classes no longer stores `stream_stats` objects (they now have overhead e.g. to store the backwards compatibility references). Statistics are now kept in std::vectors. Custom statistics are stored in `batch_stats` during a batch and merged into `stats` at the end of the batch (along with the core stats held in `add_packet_state`.


james-smith-za (approved):

james-smith-za:
Some failing checks, but it looks like it's just coverage?

bmerry:
> Some failing checks, but it looks like it's just coverage?

Yes, coveralls.io is having a major outage.

bmerry:
> Some failing checks, but it looks like it's just coverage?

Yes, coveralls.io is having a major outage.

PR #157, opened 2021-09-27
Redesign deleter for memory_allocator::pointer

bmerry:
Previously it held a shared_ptr reference to the allocator, and called
back to the allocator to free the memory via a virtual function.
This had a few disadvantages:

- The allocator had to be kept alive (and referenced) even when its
  state wasn't needed for deletion (which was pretty much all the time).
- You had to have an allocator class, which made it lots of work to
  convert a regular unique_ptr to a memory_allocator::pointer.
- Any additional state had to be crammed into a void*, which had low
  safety (raw pointer AND untyped).

Replace it all by having the deleter type derive from std::function.
There is still backwards-compatible support for custom allocators that
override `free` and/or expect to be able to call `get_allocator` and
`get_user` on the deleter, but the allocators defined by spead2 do not
use it.

### DIFF
--- a/include/spead2/common_memory_allocator.h
+++ b/include/spead2/common_memory_allocator.h
@@ -1,4 +1,4 @@
-/* Copyright 2016 National Research Foundation (SARAO)
+/* Copyright 2016, 2021 National Research Foundation (SARAO)
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU Lesser General Public License as published by the Free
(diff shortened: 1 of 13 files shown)


include/spead2/common_memory_allocator.h
```diff
@@ -37,26 +37,69 @@ class memory_allocator;
...
+
+        explicit deleter(std::shared_ptr<memory_allocator> allocator, void *user = nullptr)
+            : std::function<void(std::uint8_t *)>(legacy_deleter(std::move(allocator), user)) {}
+
+        /**
+         * If the backwards-compatibility constructor was used, return the stored
+         * allocator. Otherwise, returns a null pointer.
+         */
+        const std::shared_ptr<memory_allocator> &get_allocator() const;
+        /**
```
james-smith-za, line 98: Do these docstrings correspond to the function prototype below or above? It looks like below, but it might be slightly easier to read if there was a blank line in-between.
Just a style-thing, ignore if you disagree.

james-smith-za (approved):
Just bike-shedding on my part. Otherwise I can't really fault it.

include/spead2/common_memory_allocator.h
bmerry: It's below. I omitted the blank line because these functions are related and I wanted them visually grouped together.

PR #158, opened 2021-09-28
Make stream_stats (C++) more closely mimic STL classes

bmerry:
- Add a whole bag of typedefs
- Add `count` and `at` methods
- Add reverse iterator functions
- Make indexing operations on a const stream_stats return a const
  reference instead of a value
- Make operator[] with unknown name throw std::out_of_range rather than
  std::invalid_argument, for consistency with `at` (in a std::map it
  creates a new element, but stream_stats doesn't allow elements to be
  added or removed).

### DIFF
--- a/doc/py-recv.rst
+++ b/doc/py-recv.rst
@@ -102,7 +102,7 @@ properties after construction.
 
       Get the index of statistic `name`.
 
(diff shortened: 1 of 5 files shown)


james-smith-za (approved):
Looks fine to my untrained eye.

PR #161, opened 2021-10-08
Bump versions of things

bmerry:
Mostly so that we test against Python 3.10 and build 3.10 wheels against
the final release. But I went ahead and bumped all the versions I could
see.


bmerry:
Just adding 3.10 to the unit testing. There is still a unit test for 3.6 to make sure that it still works on older versions.

But it looks like there are still some issues with the CI. I'm going to close this and re-open it once I've got that sorted out.

PR #177, opened 2022-02-22
Update versions of various things

bmerry:
- Drop support for Python 3.6 (it's end-of-life and was making my life more difficult to support)
- Run tests against Python 3.10 in place of 3.9 (possible now that numba supports 3.10)
- Update all the versions in requirements.txt to the latest supported by Python 3.7
- Update pytest settings to support an update in pytest-asyncio
- Pin setuptools to an older version to work around a bug that prevents wheels being built
- Limit wheel building to manylinux builds (cibuildwheel now also tries to build musllinux wheels, which don't build)


james-smith-za (approved):
Seems good to me.

PR #178, opened 2022-02-24
Improve accuracy of rate limiter

bmerry:
Previously it could be a bit off because wait times were rounded to
whole nanoseconds (or whatever the unit of high_resolution_clock is).
While the actual sleeps can only be this accurate, the residual errors
would accumulate and should eventually be corrected.

This is fixed by using a new precise_time class for internal time
calculations, that holds a time_point for the high resolution clock, and
a small correction term.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -4,6 +4,9 @@ Changelog
 .. rubric:: Development version
 
 - Drop support for Python 3.6, which has reached end-of-life.
+- Improve the accuracy of the rate limiter. Previously it could send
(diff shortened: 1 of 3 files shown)


james-smith-za (approved):

james-smith-za:
Some checks failing though.

bmerry:
> Some checks failing though.

I'm poking the CI to re-run things - they seem to be timeouts and ports being in use when the test wants them, rather than any relation to the changes.

PR #184, opened 2022-04-06
Fix an asyncio.InvalidStateError

bmerry:
It occurs when the future returned by `async_send_heap` or
`async_send_heaps` is cancelled before it completes.

Also fixed another potential instance with cancelling TcpStream.connect,
although it hasn't been observed yet.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+.. rubric:: Development version
+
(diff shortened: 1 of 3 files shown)


bmerry:
@amishatishpatel could you check if this solves your InvalidStateError in NGC-536?

amishatishpatel (approved):
Checked and confirmed to be working. Thank you!

PR #185, opened 2022-06-21
Add Linux "cooked" mode support for pcap dump reader

rtobar:
Today while trying to read a pcap dump from the Tallon correlator for SKA-MID I ran into an issue with the pcap reader not being able to read these files. After some debugging I found out that this was because the data had been captured using the "any" device, as in their experiments they send some data from a server on the loopback interface (i.e, the descriptors) and some other data from the FPGAs.

After learning the difference in the headers, I came up with these two commits that add support for cooked frames to the spead2 pcap reader. The first adds the actual parsing and unit tests, while the second uses this in the pcap reader. Note that the pcap reader didn't previously check that the frames being read were ethernet frames, and thus the problem wasn't obvious at first. A check has been added now to check for the supported types.

For reference, this is the dump: https://drive.google.com/file/d/1FwUGF1d2hzoMO8FFnzsdqo_C1rxTzTq6/view?usp=sharing (requires SKA-related account access)

### DIFF
--- a/include/spead2/common_raw_packet.h
+++ b/include/spead2/common_raw_packet.h
@@ -187,6 +187,26 @@ public:
     ipv4_packet payload_ipv4() const;
 };
 
+/* Wraps a block of contiguous data and provides access to the linux sll
+ * frame header fields.
+ */
+class linux_sll_frame : public packet_buffer
+{
+public:
+    static constexpr std::size_t min_size = 16;
+
+    linux_sll_frame() = default;
+    linux_sll_frame(void *ptr, std::size_t size);
+
+    SPEAD2_DECLARE_FIELD_BE(0, std::uint16_t, packet_type)
+    SPEAD2_DECLARE_FIELD_BE(2, std::uint16_t, arphrd_type)
+    SPEAD2_DECLARE_FIELD_BE(4, std::uint16_t, link_layer_address_length)
+    SPEAD2_DECLARE_FIELD(6, std::uint64_t, link_layer_address,)
+    SPEAD2_DECLARE_FIELD_BE(14, std::uint16_t, protocol_type)
+
+    ipv4_packet payload_ipv4() const;
+};
+
 #undef SPEAD2_DECLARE_FIELD
 
 class packet_type_error : public std::runtime_error
@@ -202,6 +222,15 @@ public:
  * @throws packet_type_error if there are other problems e.g. it is not an IPv4 packet
  */
 packet_buffer udp_from_ethernet(void *ptr, size_t size);
+
+/**
+ * Inspect a Linux SLL encapsulation frame (used when listening to the "any"
+ * device) to extract the UDP4 payload, with sanity checks.
+ *
+ * @throws length_error if any length fields are invalid
+ * @throws packet_type_error if there are other problems e.g. it is not an IPv4 packet
+ */
+packet_buffer udp_from_linux_sll(void *ptr, size_t size);
 
 } // namespace spead2
 
--- a/src/common_raw_packet.cpp
+++ b/src/common_raw_packet.cpp
@@ -248,14 +248,29 @@ ipv4_packet ethernet_frame::payload_ipv4() const
     return ipv4_packet(data() + min_size, size() - min_size);
 }
 
-packet_buffer udp_from_ethernet(void *ptr, size_t size)
+/////////////////////////////////////////////////////////////////////////////
+
+linux_sll_frame::linux_sll_frame(void *ptr, std::size_t size)
+    : packet_buffer(ptr, size)
 {
-    ethernet_frame eth(ptr, size);
-    if (eth.ethertype_be() != htobe(ipv4_packet::ethertype))
+    if (size < min_size)
+        throw std::length_error("packet is to small to be a Linux sll frame");
+}
+
+ipv4_packet linux_sll_frame::payload_ipv4() const
+{
+    // TODO: handle VLAN tags
+    return ipv4_packet(data() + min_size, size() - min_size);
+}
+
+/////////////////////////////////////////////////////////////////////////////
+
+static packet_buffer udp_from_ipv4(std::uint16_t ethertype_be, const ipv4_packet &ipv4)
+{
+    if (ethertype_be != htobe(ipv4_packet::ethertype))
         throw packet_type_error("Frame has wrong ethernet type (VLAN tagging?), discarding");
     else
     {
-        ipv4_packet ipv4 = eth.payload_ipv4();
         if (ipv4.version() != 4)
             throw packet_type_error("Frame is not IPv4, discarding");
         else if (ipv4.is_fragment())
@@ -265,6 +280,18 @@ packet_buffer udp_from_ethernet(void *ptr, size_t size)
         else
             return ipv4.payload_udp().payload();
     }
+}
+
+packet_buffer udp_from_ethernet(void *ptr, size_t size)
+{
+    ethernet_frame eth(ptr, size);
+    return udp_from_ipv4(eth.ethertype_be(), eth.payload_ipv4());
+}
+
+packet_buffer udp_from_linux_sll(void *ptr, size_t size)
+{
+    linux_sll_frame frame(ptr, size);
+    return udp_from_ipv4(frame.protocol_type_be(), frame.payload_ipv4());
 }
 
 } // namespace spead2
--- a/src/recv_udp_pcap.cpp
+++ b/src/recv_udp_pcap.cpp
@@ -22,7 +22,6 @@
 #if SPEAD2_USE_PCAP
 #include <cassert>
 #include <cstdint>
-#include <pcap/dlt.h>
 #include <string>
 #include <spead2/recv_reader.h>
 #include <spead2/recv_stream.h>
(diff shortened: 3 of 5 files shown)


bmerry:
Thanks - this sounds like a useful thing to fix. Unfortunately I won't have much time to work on spead2 for the next few weeks, but I'll eventually get around to it.

src/recv_udp_pcap.cpp
```diff
@@ -59,7 +60,12 @@ void udp_pcap_file_reader::run()
                 try
                 {
                     void *bytes = const_cast<void *>((const void *) pkt_data);
-                    packet_buffer payload = udp_from_ethernet(bytes, h->len);
+                    auto payload = [bytes, h, this] () {
+                        if (linktype == 1) {
+                            return udp_from_ethernet(bytes, h->len);
+                        }
+                        return udp_from_linux_any(bytes, h->len);
+                    }();
```
bmerry, line 68: Why build a lambda and immediately call it?

Also I put all `{`s on their own lines rather than at the end of the line?

src/recv_udp_pcap.cpp
```diff
@@ -109,6 +115,11 @@ udp_pcap_file_reader::udp_pcap_file_reader(stream &owner, const std::string &fil
         throw error;
     }
     pcap_freecode(&filter);
+    // The link type used to record this file
+    linktype = pcap_datalink(handle);
+    assert(linktype != PCAP_ERROR_NOT_ACTIVATED);
+    if (linktype != 1 && linktype != 113)
```
bmerry, line 121: Can you use the constants in <pcap/dlt.h> rather than magic numbers?

bmerry (changes requested):
Thanks, this looks like good work! I've requested a few changes but nothing major.

Looking at the manual pages it seems like there is also a DLT_LINUX_SLL2 type these days. Any idea what uses that? Maybe newer kernels?

rtobar:
Thanks for the review. I've performed the requested changes.

About the newer SLL2 format, I'm not quite sure myself (this is my very first incursion into pcap). After reading some it seems like pcap offers the ability to programs to store "any" device captures in this newer SLL2 format that contains more information (e.g., https://seclists.org/tcpdump/2020/q2/39), but the default is still to use SLL, probably for backwards compatibility reasons.

src/recv_udp_pcap.cpp
rtobar: Thanks, after a quick search I hadn't been able where the constants were defined and (sadly) settled for the magic numbers. I'll use the constants of course.

src/recv_udp_pcap.cpp
rtobar: The throw-away lambda is to perform initialisation of the `payload` variable directly when declaring it, avoiding a call to the default constructor (it's a common idiom to initialise `const` variables with complex logic, which is where I've seen it more often). In 17 RVO is also guaranteed, but probably still happens anyway (haven't double-checked though), and in the worst-case scenario you get a move.

Anyway, if you feel this is too esoteric I can replace it with a simpler default initialisation + assignment.

I've moved the `{`s to their own lines. I forget this is the style in this project. Any chance on getting a clang-format config file (+ CI job) to automatically check this?

include/spead2/common_raw_packet.h
```diff
@@ -203,6 +223,23 @@ class packet_type_error : public std::runtime_error
  */
 packet_buffer udp_from_ethernet(void *ptr, size_t size);
 
+/**
+ * Inspect an ethernet frame to extract the UDP4 payload, with sanity checks.
+ *
+ * @throws length_error if any length fields are invalid
+ * @throws packet_type_error if there are other problems e.g. it is not an IPv4 packet
+ */
+packet_buffer udp_from_ethernet(const ethernet_frame &eth);
```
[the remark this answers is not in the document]
rtobar: Exactly, forgot to remove it during development.

src/common_raw_packet.cpp
```diff
@@ -248,14 +248,29 @@ ipv4_packet ethernet_frame::payload_ipv4() const
...
+
+ipv4_packet linux_cooked_frame::payload_ipv4() const
+{
+    // TODO: handle VLAN tags
+    return ipv4_packet(data() + min_size, size() - min_size);
+}
+
+/////////////////////////////////////////////////////////////////////////////
+
+static packet_buffer udp_from_ethernet(std::uint16_t ethertype_be, const ipv4_packet &ipv4)
```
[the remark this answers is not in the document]
rtobar: Yes, I think you're right, should have thought about the naming better, will fix.

src/recv_udp_pcap.cpp
bmerry: Interesting - I'd not come across that before, although I've done similar things just using a named function rather than a lambda (usually for constructor initializer lists). But I think in this case you'd get all the same benefits just by using a ternary:

```c++
auto payload = (linktype == DL_EN10MB) ? udp_from_ethernet(bytes, h->len) : udp_from_linux_any(bytes, h->len);
```

> Any chance on getting a clang-format config file (+ CI job) to automatically check this?

I haven't used clang-format before, but my experience with code formatters in other languages (`black`, `rustfmt`) has left me with mixed feelings. For a project with multiple regular contributors I tend to favour them to enforce uniformity, but I also find they can make code less readable and sometimes even less uniform by mindless applying rules without appreciation of aesthetics. So my inclination is to say no, but if it's possible to configure clang-format with a really minimal set of rules (e.g. just the brace placement) so that it doesn't try to make hundreds of changes to the existing code I could look at it.

bmerry:
I guess we can leave SLL2 until someone asks for it. Maybe name the function `udp_from_linux_any` to `udp_from_linux_sll` to be clear about exactly which data-link type it handles.

include/spead2/common_raw_packet.h
```diff
@@ -187,6 +187,26 @@ class ethernet_frame : public packet_buffer
     ipv4_packet payload_ipv4() const;
 };
 
+/* Wraps a block of contiguous data and provides access to the linux "cooked"
+ * frame header fields.
+ */
+class linux_cooked_frame : public packet_buffer
```
bmerry, line 193: Let's call this `linux_sll_frame` to match the naming used by pcap and avoid any confusion with SLL2, which is also "cooked". And similarly update other uses of "cooked" in code/variable names.

bmerry (changes requested):

src/recv_udp_pcap.cpp
rtobar: I hadn't thought about using a ternary, thanks for the suggestion. I did think separately just now that we can avoid the branch altogether if I store a pointer to the correct function instead of checking `linkttype` each time, so I've implemented that. Let me know what you think.

I do agree with your comments about automatic formatters, I also have the same appreciation, so I understand not feling inclined to add one. Last time I checked clang-format it didn't allow enforcing single rules, which makes it even less likely to become adopted here.

include/spead2/common_raw_packet.h
rtobar: Done (to the best of my knowledge I didn't leave any trace of "cooked" anywhere)

bmerry (approved):
Thanks, looks good! Although for future reference, rebasing and force-pushing makes reviewing more difficult since I can't see what's changed between revisions.

bmerry:
Darn, I was all ready to merge, but it looks like the old version of pcap used for cibuildwheel doesn't provide pcap/dlt.h. So it will probably be necessary to define the magic numbers yourself. But I'd suggest defining macros near the top of the file, something like (untested)
```c++
// These are defined here rather than using pcap/dlt.h because old versions of pcap don't provide it
#ifndef DLT_EN10MB
#define DLT_EN10MB 1
#endif
#ifndef DLT_LINUX_SLL
#define DLT_LINUX_SLL 113
#endif
```

rtobar:
Sorry about the force pushes, I was prioritising self-contained commits that did "the right thing" instead of a history of smaller patches. I do appreciate this can make reviews harder though, so I'll stay away from force pushing again.

I've landed the requested final patch, the breaking change was introduced in libpcap 1.8.0: https://github.com/the-tcpdump-group/libpcap/commit/38d2b61d754c626f591a5abf3ef1f06339573d3e

bmerry:
Thanks. Looks good, and thanks for doing the research to figure out when libpcap changed. Assuming this passes CI I'll merge.

rtobar:
Thanks @bmerry for the thorough review as usual.

While I'm here, I wanted to ask for some advice. These data dumps I've been working with contain packets for heaps originally sent to multiple streams (listening on multiple ports). What we need to do is read these heaps and act as a sender, replicating the individual streams. What I was originally planning to do was to use a pcap reader to get the heaps, then send those with my senders to their endpoint. However, the pcap receiver currently cannot be constrained to consider only heaps belonging to a particular stream, or sent to a particular port (and there's no way to infer that information from a heap AFAIR).

Given that understanding, I was planning on adding an option to the udp_pcap reader so that it only considers UDP packets for sent to a specific port; that way I could open N receivers each with a pcap reader for each of the ports I'm interested in, then do my heap forwarding. This would be a simple change I believe, but it means I'd have to read the dump N times. An alternative implementation could feed the N receivers using a single pass over the data dump, but that would be more work to get right I believe.

So:

* Is my current understanding correct that what I want to achieve is not possible, or am I missing something?
* What do you think about the proposals stated above, and would you be open to accept patches in that direction? This would bubble up to the python bindings, although I have the intention in the longer term to move our code to C++ now that's a bit more stable.

bmerry:
@rtobar I think I've got Discussions enabled now (I've just left it with the default categories). If you repost your question there I'll take a look.

PR #187, opened 2022-07-05
Make CI builds work on MacOS again

bmerry:
Set CPPFLAGS and LDFLAGS to access /usr/local/include and /usr/local/lib. Some recent ecosystem change has prevented those paths from being picked up.

Also bump all the actions to the most recent versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,7 +106,7 @@ jobs:
         env:
           CC: ${{ matrix.cc }}  # Do not pass -Werror when building dependencies
       - run: ./bootstrap.sh
-      - run: pip install -v .
+      - run: CPPFLAGS="-I/usr/local/include" pip install -v .
       - name: Run tests
         run: ./.ci/py-tests.sh
       - run: flake8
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -98,15 +98,18 @@ jobs:
...
       - name: Install Python dependencies
         run: ./.ci/py-requirements.sh
         env:
           CC: ${{ matrix.cc }}  # Do not pass -Werror when building dependencies
       - run: ./bootstrap.sh
-      - run: pip install -v .
+        # CPPFLAGS and LDFLAGS are to work around Python config oddities that
+        # cause it to run clang with options that prevent these locations being
+        # searched.
+      - run: CPPFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib" pip install -v .
```
ludwigschwardt, line 112: `/usr/local` is actually the output of `brew --prefix` but this should be safe on the GitHub Action runners.

ludwigschwardt, line 112: `/usr/local` is actually the output of `brew --prefix` but this should be safe on the GitHub Action runners. And we probably want a cleaner solution eventually, so this is fine for now.

ludwigschwardt (approved):

PR #191, opened 2022-07-18
Add support for user-provided libpcap filter

rtobar:
This PR adds support for user-provided filters when adding a new pcap reader as discussed in #189. This in turn enables usage of capture files containing more than a single stream.

My only concern here is the usage of `filter` as argument name in the python function, as it overrides the `filter` built-in function. However, since the implementation of the `add_udp_pcap_file_reader` python is done in C++ anyway it should never become a problem. I'm more than open otherwise to other names.


bmerry:
Thanks. From the quick glance it looks good, but I'll need to find time to review it properly (spead2 currently isn't high on my priority list).

> My only concern here is the usage of filter as argument name in the python function, as it overrides the filter built-in function.

I wouldn't worry about it: parameter names don't clash with function names except when implementing the function, and I've seen (and written) plenty of code with names like `filter` and `type` before.

src/recv_udp_pcap.cpp
```diff
@@ -108,8 +108,11 @@ udp_pcap_file_reader::udp_pcap_file_reader(stream &owner, const std::string &fil
         throw std::runtime_error(errbuf);
     // Set a filter to ensure that we only get UDP4 packets with no fragmentation
     bpf_program filter;
+    std::string filter_expression = "ip proto \\udp and ip[6:2] & 0x3fff = 0";
+    if (!user_filter.empty())
+        filter_expression += " and (" + user_filter + ')';
     if (pcap_compile(handle, &filter,
-                     "ip proto \\udp and ip[6:2] & 0x3fff = 0",
+                     filter_expression.data(),
```
bmerry, line 115:
```suggestion
                     filter_expression.c_str(),
```
I know they're the same since C++11, but subconsciously I still look at `.data()` and think of it as not being null-terminated.

PR #192, opened 2022-07-21
Add experimental support for a shared library

bmerry:


bmerry:
@ludwigschwardt would you mind testing this out on a Mac? Try building with `configure --enable-shared` and check that you get a Mac shared library out.

ludwigschwardt:
I removed all the build fluff and ran this on my laptop:
```
./bootstrap.sh
configure --enable-shared
make
```
This resulted in a dylib as such (single architecture):
```
git[shared-lib]$ file src/.libs/libspead2.1.dylib 
src/.libs/libspead2.1.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```
I could not test what happens with `setup.py` in this case.

bmerry:
Thanks, that's what I was hoping to get.

> I could not test what happens with setup.py in this case.

This is aimed at users of the C++ interface, so that's not relevant.

PR #193, opened 2022-08-05
Upgrade MacOS version used in Github Actions

bmerry:


.github/workflows/test.yml
```diff
@@ -106,10 +106,7 @@ jobs:
         env:
           CC: ${{ matrix.cc }}  # Do not pass -Werror when building dependencies
       - run: ./bootstrap.sh
-        # CPPFLAGS and LDFLAGS are to work around Python config oddities that
-        # cause it to run clang with options that prevent these locations being
-        # searched.
-      - run: CPPFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib" pip install -v .
```
ludwigschwardt, line 112: Did this just start working all of a sudden without the extra flags? :-) I'm still on macOS 11 and for me sometimes it works and sometimes it doesn't.

ludwigschwardt (approved):
Interesting about the build flags...

.github/workflows/test.yml
bmerry: It seemed to, and the builds no longer seem to be passing `-isysroot`.

PR #199, opened 2022-08-05
Move most of setup.py into setup.cfg

bmerry:
This is a bit of maintenance which will allow other tools (hopefully
including dependabot) to access metadata.

It also makes the code a bit simpler, because we can use the `file:` and
`attr:` magic in setup.cfg.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -141,7 +141,7 @@ jobs:
       - run: ./bootstrap.sh
       - name: Prepare Python build for coverage
         run: |
-          echo '[build_ext]' > setup.cfg
+          echo '[build_ext]' >> setup.cfg
           echo 'coverage = yes' >> setup.cfg
       - run: pip install -v .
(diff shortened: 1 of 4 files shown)


ludwigschwardt (approved):
Nice! I'm going to apply this elsewhere, I hope :-)

bmerry:
Unfortunately it doesn't seem like there is a nice clean way to tell Dependabot what Python version to use when updating your dependencies (just [yucky poorly-documented ones](https://github.com/dependabot/dependabot-core/issues/1455)).

ludwigschwardt:
Yeah, I hear you (isn't `.python-version` meant for pyenv?).

bmerry:
> isn't .python-version meant for pyenv?

Precisely what other people have raised. But also, I wasn't fully aware of pyenv and pyenv-virtualenv, so I'm experimenting with it now and liking the fact that it can automatically switch virtualenvs when I change directories.

PR #200, opened 2022-08-11
Mark chunking receiver as no longer experimental

bmerry:


james-smith-za (approved):
I guess it's been pretty stable for the last few months.

PR #203, opened 2022-09-07
Use autodoc's mocking mechanism instead of our own

rtobar:
The latter prevented asyncio.TcpStream.connect's documentation from being generated, which was confusing. Using autodoc's own mocking mechanism seems to fix this issue, and hopefully prevent future ones.

See https://spead2.readthedocs.io/en/latest/py-send.html#id2 for the current missing documentation.

Signed-off-by: Rodrigo Tobar <rtobar@icrar.org>

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -25,6 +25,8 @@ sys.path.insert(0, os.path.abspath('..'))
 rtd = os.environ.get('READTHEDOCS') == 'True'
 if rtd:
     subprocess.check_call(['doxygen'])
+    # ReadTheDocs can't build the extension (no Boost), so we have to mock it
+    # See http://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
     autodoc_mock_imports = ['spead2._spead2', 'spead2._spead2.send', 'spead2._spead2.recv']
 
 # -- Project information -----------------------------------------------------
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -27,7 +27,7 @@ if rtd:
     subprocess.check_call(['doxygen'])
     # ReadTheDocs can't build the extension (no Boost), so we have to mock it
     # See http://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
-    autodoc_mock_imports = ['spead2._spead2', 'spead2._spead2.send', 'spead2._spead2.recv']
+    autodoc_mock_imports = ['spead2._spead2']
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


bmerry:
Thanks, I didn't know this feature existed! It looks like it was added around the same time I started writing docs for spead2 and I think readthedocs was still recommending the manual process I used, and ever since I've just cargo-culted it into other projects.

When I get a chance I'll check that nothing else goes missing as a result.

doc/conf.py
```diff
@@ -25,14 +25,7 @@
...
     subprocess.check_call(['doxygen'])
-    # ReadTheDocs can't build the extension (no Boost), so we have to mock it
-    # See http://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
-    import unittest.mock
-    MOCK_MODULES = ['spead2._spead2', 'spead2._spead2.send', 'spead2._spead2.recv']
-    sys.modules.update((mod_name, unittest.mock.Mock()) for mod_name in MOCK_MODULES)
-    # Mocking certain classes causes subclasses not to be documented properly
-    sys.modules['spead2._spead2.send'].UdpStreamAsyncio = object
-    sys.modules['spead2._spead2.send'].UdpIbvStreamAsyncio = object
+    autodoc_mock_imports = ['spead2._spead2', 'spead2._spead2.send', 'spead2._spead2.recv']
```
bmerry, line 28: According to the docs, it's not necessary to list submodules, so I believe this can be reduced to:
```suggestion
    autodoc_mock_imports = ['spead2._spead2']
```

doc/conf.py
```diff
@@ -25,14 +25,7 @@
 rtd = os.environ.get('READTHEDOCS') == 'True'
 if rtd:
     subprocess.check_call(['doxygen'])
-    # ReadTheDocs can't build the extension (no Boost), so we have to mock it
-    # See http://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
```
bmerry, line 29: Can you preserve this comment.

bmerry (changes requested):
Thanks, it looks like this also fixes documentation of a couple of other classes/functions too.

doc/conf.py
rtobar: Ahhh true! I knew this, but did the change in the hurry and didn't stop to think about it, good catch.

I also learned about autodoc's mock mechanism after doing my own mocking for some time. I've run into some small gotchas, but in almost all cases it does a pretty good work.

doc/conf.py
rtobar: done

doc/conf.py
rtobar: I've applied the requested change

PR #206, opened 2022-09-27
Support placing extra data into chunks

bmerry:
Allow the place callback to emit a block of data to be placed into the chunk at a particular offset. This makes it practical to capture immediate items from the heaps.


james-smith-za (approved):
Looks fine to me, though I'll frankly admit that the C++ is at the level that I can't really add any value by looking at it.

PR #215, opened 2023-03-30
Update unit tests where async tasks were unhandled

amishatishpatel:
Just the two unit tests, actually.

Contributes to: NGC-899.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+.. rubric:: Development version
+
+- Update test_send_asyncio.py's :py:meth:`test_async_flush` and
+  :py:meth:`test_async_flush_fail` to keep handles on its async Tasks.
+
 .. rubric:: 3.11.1
 
 - Fix a packaging issue that meant automake and similar tools were required to
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,8 +3,8 @@ Changelog
 
 .. rubric:: Development version
 
-- Update test_send_asyncio.py's :py:meth:`test_async_flush` and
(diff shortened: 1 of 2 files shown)


doc/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+.. rubric:: Development version
+
+- Update :program:`test_send_asyncio.py`'s ``test_async_flush`` and
+  ``test_async_flush_fail`` to keep handles on its async Tasks.
```
amishatishpatel, line 7: I took inspiration on formatting from the last mention of a file and unit test further down (3.7.0 and 1.13.0).

bmerry, line 7: test_send_asyncio.py isn't a "program". Here's my suggestion (the `!` tells Sphinx not to try creating a link, which won't be found since the tests don't have docstrings):
```suggestion
- Update :meth:`!test_async_flush` and :meth:`!test_async_flush_fail` to keep
  handles to async tasks, to prevent them being garbage collected too early.
```

bmerry (changes requested):
Code looks good, just nitpicking the changelog entry.

doc/changelog.rst
bmerry: Ah, hadn't seen that. `:program:` is for an executable, but I guess it would be fine to use the double backticks for the test names for consistency.

james-smith-za (approved):
Other than Bruce's bike-shedding, looks OK to me.

bmerry (approved):

bmerry:
Looks good. Let's just wait for all the CI tests to pass, then you can merge.

PR #219, opened 2023-07-05
Add chunk stream groups

bmerry:
This allows multiple streams to work on assembling a shared chunk, giving better scalability provided the incoming data can be distributed across streams. A good starting point for review is doc/recv-chunk-group.rst.

This is quite a large PR, because it also incorporates a change in the way streams are shut down.

I'm not totally happy with the amount of testing, but I'm starting the ball rolling, and I can always add more tests before release.


doc/changelog.rst
```diff
@@ -5,6 +5,16 @@ Changelog
...
+- Fix a bug where copying a :cpp:class:`spead2::recv::stream_config` would not
+  deep copy the names of custom statistics, and so any statistics added to the
+  copy would also affect the original, and there were also potential race
+  conditions if a stream config was modified while holding stream statistics.
+- Fix a bug (caused by the bug above) where passing a
+  :cpp:class:`spead2::recv::stream_config` to construct a
+  :cpp:class:`spead2::recv::chunk_stream` would modify the config. Passing
+  the same config to construct two chunk streams would fail with an error.
+- Fix the type annotation for the :py:class:`~.ChunkRingStream` constructor:
+  the parameter name for `chunk_stream_config` was incorrect.
```
bmerry, line 17: I guess I'll need to update the changelog for chunk stream groups and the new shutdown method.

james-smith-za (approved):
In all honesty I don't have a lot to contribute here apart from a little spelling / punctuation bike-shedding.

I am not familiar enough (or at all really) with the internals of spead2 in order to be able to comment on the nitty gritty.
What I will say though was that the user-level documentation was pretty clear, and the examples were easy enough to follow. My C++ is good enough to follow that one at least.

PR #223, opened 2023-07-11
Fix a shutdown hang

bmerry:
If the process exits without explicitly stopping anything, it's possible deadlock if the threadpools used for the streams in a group were created after the group. This happens because `atexit` stops the threadpools before it stops the group.

Fix it by giving the exit_stopper treatment to
chunk_stream_group_member. Since the member is necessarily constructed after the thread pool it uses, this ensures that the member (and hence the whole group) is stopped before the thread pool.

### DIFF
--- a/tests/shutdown.py
+++ b/tests/shutdown.py
@@ -1,4 +1,4 @@
-# Copyright 2017-2018 National Research Foundation (SARAO)
+# Copyright 2017-2018, 2023 National Research Foundation (SARAO)
 #
 # This program is free software: you can redistribute it and/or modify it under
 # the terms of the GNU Lesser General Public License as published by the Free
@@ -15,16 +15,23 @@
 
 """Tests for shutdown ordering.
 
-These cannot be run through nosetests because they deal with interpreter
+These cannot be run through pytest because they deal with interpreter
 shutdown.
 """
 
 import logging
 
+import numba
+import numpy as np
+import scipy
+from numba import types
+
 import spead2
 import spead2.recv
 import spead2.send
 import spead2._spead2
+from spead2.numba import intp_to_voidptr
+from spead2.recv.numba import chunk_place_data
 
 
 def test_logging_shutdown():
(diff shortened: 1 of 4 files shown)


james-smith-za (approved):

james-smith-za:
Is there potentially a unit test that you could think of to catch this behaviour? How did you find it?

bmerry:
> Is there potentially a unit test that you could think of to catch this behaviour? How did you find it?

I found it by hitting Ctrl-C on a script I was running. It's something that happens during the interpreter shutdown process, so not amenable to pytest, but there are actually some tests for this sort of thing. Let me add something.

tests/shutdown.py
```diff
@@ -55,3 +62,53 @@ def test_running_stream():
...
+
+    # Send 4 chunks of data to the first stream. This will cause it to block
+    # while it waits for the second stream to progress.
+    send_stream = spead2.send.InprocStream(spead2.ThreadPool(), queues)
+    ig = spead2.send.ItemGroup()
+    ig.add_item(0x1000, "payload", "payload", shape=(1024,), dtype=np.uint8)
+    ig["payload"].value = np.ones(1024, np.uint8)
+    heap = ig.get_heap(descriptors="none", data="all")
+    for _ in range(4):
+        send_stream.send_heap(heap, substream_index=0)
```
james-smith-za, line 114: I'm not quite understanding what the expected outcome of this is.

I tried installing spead2 on my workstation so I could run the tests, but ... I couldn't get it installed.

bmerry: It puts the streams into a state where something is blocked (specifically, the first stream has received a bunch of chunks, but since the group has `max_chunks=2` it can't acquire the 3rd chunk until the first chunk is aged out, and that can't happen until the second stream receives the data for it, which never happens). Then the process exits, and the test is that the process actually does terminate rather just hanging waiting for progress. If I remove the fix made in this PR, then running this test just hangs.

james-smith-za: Right, I got you.
If I *really* wanted to nit-pick, I'd ask for a comment or something at the bottom indicating that we just expect the process to exit rather than to block. In case at some future point the tests do hang here, and the intrepid investigator then can't figure out why.

bmerry: Well more generally it tests that the shutdown is clean; segfaulting is another possible failure mode. And these tests are marked as "shutdown" tests.

PR #227, opened 2023-07-13
Use pybind11's setup helpers

bmerry:
The main benefit is that it has tools to support parallel compilation,
which drastically cuts down the build time. It also determines the
latest version of C++ supported by the compiler and uses that, which in
a few cases lets pybind11 do things more efficiently. Finally, it takes
care of some of the boilerplate, like passing -g0 -fvisibility=hidden.


james-smith-za (approved):
I like.

PR #234, opened 2023-08-07
Add detection of function multi-versioning in the compiler

bmerry:
This support is used so that the SSE2 version of memcpy_nontemporal is selected only if the CPU supports SSE2 at runtime.

On x86-64 there is guaranteed to be runtime support, so this only affects i386. Note that i386 is not currently tested. The real motivation is that it lays the groundwork for adding more specialised variants e.g. AVX, AVX-512, MOVDIR64B etc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,6 +155,20 @@ SPEAD2_ARG_WITH(
     )]
 )
 
+# -Werror is used to get an error (rather than a warning) if the compiler
+# doesn't support the feature.
+SPEAD2_ARG_WITH(
+    [fmv],
+    [AS_HELP_STRING([--without-fmv], [Do not use compiler function multi-versioning support])],
+    [SPEAD2_USE_FMV],
+    [SPEAD2_CHECK_FEATURE(
+        [fmv], [function multi-versioning], [], [], [],
+        [SPEAD2_USE_FMV=1], [],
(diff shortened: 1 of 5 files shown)


bmerry:
@james-smith-za I think hold off the reviewing for now - it's going to need some tweaks to make it play nice with MacOS's universal binaries.

james-smith-za:
> @james-smith-za I think hold off the reviewing for now - it's going to need some tweaks to make it play nice with MacOS's universal binaries.

Why we are bothering to support macOS is still beyond me...

Does the Apple silicon even support SSE2?

bmerry:
> Why we are bothering to support macOS is still beyond me...

Some applications that use spead2 are developed by people who have Macs (like @ludwigschwardt).

> Does the Apple silicon even support SSE2?

It doesn't. But the Python builder tries to build universal binaries that have code for both x86-64 and ARM64, which confuses things.

bmerry:
@james-smith-za looks like I've got the MacOS support fixed.

rtobar:
> > Why we are bothering to support macOS is still beyond me...
> 
> Some applications that use spead2 are developed by people who have Macs (like @ludwigschwardt).

To further support this point: in SKA there are several developers that use Macs and either indirectly use spead2, or actually develop applications on top of it.

james-smith-za (approved):

PR #237, opened 2023-08-17
Have cibuildwheel test the wheels it builds

bmerry:
This improves test coverage in a few ways:
- it tests against all the Python versions for which wheels are built (the rest of the unit tests only cover min and max versions)
- it tests the wheels rather than the source code
- once other architectures are added, those will get tests too

### DIFF
--- a/tests/test_recv.py
+++ b/tests/test_recv.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import platform
 import socket
 import struct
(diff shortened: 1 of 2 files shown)


james-smith-za (approved):

tests/test_recv.py
```diff
@@ -1231,7 +1232,8 @@ def test_out_of_range_udp_port(self):
             receiver.add_udp_reader(100000)
 
     @pytest.mark.skipif(
-        sys.platform == "darwin", reason="Test does not work with macos on Github Actions"
+        sys.platform == "darwin" or os.environ.get("CIBUILDWHEEL") == "1",
+        reason="Test does not work with macos on Github Actions or under cibuildwheel",
```
rtobar, line 1236: While browsing this, I thought I would comment that we have a similar test in a different, but instead of using a privileged port to trigger an error (seems to be the rationale here) we use an out-of-range value (e.g., 100000). After all, 22 isn't an illegal UDP port

bmerry: That test is in here too (just a few lines further up). They're testing different failure modes though. One is an error when trying to pass a large number from Python to a uint16_t in C++, while the other is an error coming from the operating system.

rtobar: I see, I didn't explore much further than what the GitHub diff UI showed me, sorry for the noise

PR #239, opened 2023-08-17
Update to require at least C++17

bmerry:
There are many simplifications made possible by requiring C++17, and some optimisations (mostly from C++14 lamba initialisers allowing data to be moved rather than copied into lambdas). Apart from those, there are a few more major changes:
- common_bind.h is deleted. It wasn't being used anywhere.
- a new `storage` class is introduced to provide a safer alternative to `std::aligned_storage` (which is deprecated in C++23)
- testing now covers a wider range of compiler versions


bmerry:
@sjperkins the Github interface only seems to let me request reviews from certain people so I can't tag you for review, but please take a look when convenient.

@rtobar @ajameson let me know if you hit any compiler issues with this branch.

rtobar:
Compiles here (standalone and python package) with g++ 12.2.0 on Ubuntu 22.10.

I had other compilation woes though, will report separately.

include/spead2/py_common.h
```diff
@@ -392,105 +394,50 @@ namespace detail
...
-PTMFWrapperGen<T, Return, Class, Args...> ptmf_wrapper_type(Return (Class::*ptmf)(Args...) const);
-
-#define SPEAD2_PTMF(Class, Func) \
-    (decltype(::spead2::detail::ptmf_wrapper_type<Class>(&Class::Func))::template make_wrapper<&Class::Func>())
-// Discards the return value - useful for setters where the C++ function returns *this
-#define SPEAD2_PTMF_VOID(Class, Func) \
-    (decltype(::spead2::detail::ptmf_wrapper_type<Class>(&Class::Func))::template make_wrapper_void<&Class::Func>())
+#define SPEAD2_PTMF(Class, Func) (::spead2::detail::make_ptmf<Class, &Class::Func>(&Class::Func))
+// Like the above, but additionally ignore the return value
+#define SPEAD2_PTMF_VOID(Class, Func) (::spead2::detail::make_ptmf_void<Class, &Class::Func>(&Class::Func))
```
sjperkins, line 440: Seems like it was possible to remove a lot of code here

include/spead2/recv_chunk_stream_group.h
```diff
@@ -198,28 +196,28 @@ class chunk_stream_group
     class dereference
     {
     public:
-        decltype(*std::declval<T>()) operator()(const T &ptr) const { return *ptr; }
+        decltype(auto) operator()(const T &ptr) const { return *ptr; }
```
sjperkins, line 199: Is it possible to just have an `auto` return type here (and for the `derefence_const` lower down?

src/spead2_send.cpp
```diff
@@ -202,8 +203,8 @@ sender::sender(const options &opts)
         pointers[i].resize(opts.items);
         for (std::size_t j = 0; j < opts.items; j++)
         {
-            pointers[i][j] = reinterpret_cast<element_t *>(
-                storage.get() + i * heap_size + j * item_size);
+            pointers[i][j] = std::launder(reinterpret_cast<element_t *>(
```
sjperkins, line 206: Also just to be sure I guess.

include/spead2/recv_chunk_stream.h
```diff
@@ -490,7 +488,7 @@ class chunk_manager_simple
     std::uint64_t *get_batch_stats(chunk_stream_state<chunk_manager_simple> &state) const;
     chunk *allocate_chunk(chunk_stream_state<chunk_manager_simple> &state, std::int64_t chunk_id);
     void ready_chunk(chunk_stream_state<chunk_manager_simple> &state, chunk *c);
-    void head_updated(chunk_stream_state<chunk_manager_simple> &state, std::uint64_t head_chunk) {}
+    void head_updated(chunk_stream_state<chunk_manager_simple> &, std::uint64_t) {}
```
sjperkins, line 491: `[[maybe_unused]]`?

sjperkins (commented):
I had a look through and I think the only thing I can think that needs to change in this PR is making a hard-coded packetsize constant `static constexpr`. Otherwise, the following set of general changes is implemented: 

- `stream_base::queue_storage` changes from `storage_type` to `queue_entry`.
- `spead2::detail::storage` is a vendored `std::aligned_storage` with a cleaner API.
- `std::make_unique` used instead of `std::unique_ptr<T>(new T)`
- `std::launder` used to avoid unwanted compiler optimisations on placement new allocated types.
- Removes some callable wrapper classes in favour of c++14 lambdas
- `[[maybe_unused]]` to remove compiler warnings for unused arguments. In other places, arguments are anonymised. I wondered why there is a distinction between the two cases? Possibly for the sake of existing API calls?

src/recv_tcp.cpp
```diff
@@ -120,7 +114,7 @@ void tcp_reader::packet_handler(
 bool tcp_reader::parse_packet(stream_base::add_packet_state &state)
 {
     assert(pkt_size > 0);
-    assert(tail - head >= pkt_size);
+    assert(tail - head >= std::ptrdiff_t(pkt_size));
```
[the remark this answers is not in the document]
bmerry: Yes

src/recv_packet.cpp
```diff
@@ -40,8 +38,8 @@ namespace recv
 template<typename T>
 static inline T extract_bits(T value, int first, int cnt)
 {
-    assert(0 <= first && first + cnt <= 8 * sizeof(T));
-    assert(cnt > 0 && cnt < 8 * sizeof(T));
+    assert(0 <= first && first + cnt <= 8 * int(sizeof(T)));
+    assert(cnt > 0 && cnt < 8 * int(sizeof(T)));
```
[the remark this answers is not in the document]
bmerry: Yes

src/spead2_recv.cpp
```diff
@@ -226,7 +227,7 @@ static void verify_heap(const spead2::recv::heap &fheap, const options &opts)
                 << " has an inconsistent length\n";
             std::exit(1);
         }
-        const element_t *data = reinterpret_cast<const element_t *>(item.ptr);
+        const element_t *data = std::launder(reinterpret_cast<const element_t *>(item.ptr));
```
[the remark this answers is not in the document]
bmerry: The memory is allocated by some allocator function (probably make_unique_for_overwrite) and filled with bytes from the network. I don't fully understand std::launder, but I think this is a case where it applies since that allocation "provides storage" for the array of element_t.

include/spead2/py_common.h
```diff
@@ -561,28 +508,6 @@ struct type_caster<spead2::socket_wrapper<SocketType>>
...
-        else if (src.is_none())
-            return true;
-        make_caster<spead2::socket_wrapper<SocketType>> inner_caster;
-        if (!inner_caster.load(src, convert))
-            return false;
-        value = cast_op<spead2::socket_wrapper<SocketType> &&>(std::move(inner_caster));
-        return true;
-    }
-};
-
```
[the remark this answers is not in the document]
bmerry: Correct.

include/spead2/recv_chunk_stream.h
bmerry: I went with anonymising where the function body is empty, since in that case the function is clearly just there to satisfy a concept and you're unlikely to want to refer to the parameter in future. Where there is a function body that happens not to (currently) refer to the parameter I used `[[maybe_unused]]`.

sjperkins (approved):

PR #241, opened 2023-08-18
Build aarch64 wheels

bmerry:
To speed things up, the minimum Boost version is now 1.69 (close to 5 years old, so should be available in most non-EOL OSes already) and the Python build no longer requires any compiled Boost libraries (just headers).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -194,6 +194,9 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
+      - uses: docker/setup-qemu-action@v2
+        with:
+          platforms: arm64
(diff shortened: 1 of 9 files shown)


james-smith-za (commented):
Seems fine to me, but cibuildwheel appears to be failing.

bmerry:
The symptoms look like what happens if /etc/protocols is missing - I'll investigate.

bmerry:
The cibuildwheel failures seem to be because qemu doesn't support one of the socket options. I've opened a [ticket](https://gitlab.com/qemu-project/qemu/-/issues/1837), but for now I'll probably need to modify the test to skip it if that socket option doesn't function.

james-smith-za (approved):

bmerry:
Thanks. Guess we'll see if the wheels successfully build this time.

PR #242, opened 2023-08-22
Move mypy configuration into pyproject.toml

bmerry:
This keeps all the static checkers (isort, black, mypy) together, except for flake8 which refuses to use pyproject.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,0 @@
-[mypy]
-ignore_missing_imports = 1
-files = src/spead2, examples, tests
-python_version = 3.8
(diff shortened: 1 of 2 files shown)


james-smith-za (approved):
ruff has near-parity with flake8 and uses pyproject.toml, IIRC.

Just a thought, use it, don't use it...

bmerry:
> ruff has near-parity with flake8 and uses pyproject.toml, IIRC.

Hmm, tempting actually. Not that I particularly need flake8 to be fast - it's mypy that takes the time in pre-commit.

james-smith-za:
The other advantage is that ruff implements the functionality of isort and pydocstyle (according to [this](https://github.com/astral-sh/ruff#rules) ). In time it may end up able to replace black as well.

Probably no action required for the time being but it's something to keep in mind for later on if you're bored. I've used it a little bit but I can't say I've done an extensive comparison with what we currently use.

PR #243, opened 2023-08-23
Replace flake8 with ruff

bmerry:
I also enabled a few extra plugins beyond the flake8 basics, and fixed resulting errors.


src/spead2/__init__.pyi
```diff
@@ -112,7 +98,7 @@ class MemoryPool(MemoryAllocator):
         upper: int,
         max_free: int,
         initial: int,
-        allocator: Optional[MemoryAllocator] = None,
+        allocator: MemoryAllocator | None = None,
```
james-smith-za, line 101: My only concern is that this kind of change will break backwards compatibility. Unless for new versions of spead2 you are only planning on supporting Python 3.10+?

james-smith-za: Ah, that makes sense.

james-smith-za (approved):

PR #249, opened 2023-08-28
Fix detection of gdrapi

bmerry:
It was using the C++ compiler for detection, but the CUDA compiler when
building. They have different options to configure them (cxx_flags
versus cuda_flags).

The implementation is more complex than I would have liked, because the
cuda compiler support in meson doesn't implement has_function.


james-smith-za (approved):
Seems sensible to me.

PR #258, opened 2023-08-31
Create wheels for MacOS

bmerry:


james-smith-za (approved):

ludwigschwardt (approved):
A coveralls regression is acceptable in my books :-)

bmerry:
> A coveralls regression is acceptable in my books :-)

Yes, it goes up and down randomly, I think because of some lock contention.

PR #264, opened 2023-09-04
Use a more generic workaround for pytest-dev/pytest#11374

bmerry:
Instead of selectively clearing members in teardown methods, just clear the whole object every time using a pytest hook.


james-smith-za (approved):
LGTM

PR #265, opened 2023-09-05
Start of developer guide

bmerry:
Add basic instructions for new developers to get a development environment set up and ready to make contributions.


doc/dev-setup.rst
```diff
@@ -0,0 +1,159 @@
...
+annotations, because annotations in the :file:`.pyi` files take precedence
+(and spead2 pre-dates Python 3 annotation syntax). New code (particularly in
+tests) can be annotated, but it is not required.
+
+Identifiers use US English spelling, but comments, log messages and
+documentation favour UK spelling.
+
+The C++ code is less consistent in style, but here are some guidelines:
+
+- Use 4 spaces for indentation (**never** tabs).
```
james-smith-za, line 90: Amen brother.

doc/dev-setup.rst
```diff
@@ -0,0 +1,159 @@
...
+- Opening braces go on their own line (Allman style). An exception is that a
+  function may be written entirely on one line if it is very short.
```
james-smith-za, line 92: Preach!

This is the one thing that bothers me about Go syntax, requiring the { not to be on its own line.

james-smith-za (commented):
Everything looks fine, but I ran into an issue when I was following the instructions myself. I did it on dbelab07 just to ensure a fresh environment.

```

(spead2) jsmith@dbelab07:~/spead2$ pip install --no-build-isolation -e .
Obtaining file:///home/jsmith/spead2
  Checking if build backend supports build_editable ... done
  Preparing editable metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing editable metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [60 lines of output]
      + meson setup /home/jsmith/spead2 /home/jsmith/spead2/build/cp310 -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dtools=disabled -Dunit_test=disabled -Dcuda=disabled -Dgdrapi=disabled -Dcap=disabled -Dpython=true --native-file=/home/jsmith/spead2/build/cp310/meson-python-native-file.ini
      The Meson build system
      Version: 1.2.1
      Source dir: /home/jsmith/spead2
      Build dir: /home/jsmith/spead2/build/cp310
      Build type: native build
      Project name: spead2
      Project version: 4.0.0
      C++ compiler for the host machine: ccache c++ (gcc 11.4.0 "c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0")
      C++ linker for the host machine: c++ ld.bfd 2.38
      Host machine cpu family: x86_64
      Host machine cpu: x86_64
      Program python (jinja2, pycparser, packaging) found: YES (/home/jsmith/spead2/.venv/bin/python3) modules: jinja2, pycparser, packaging
      Found pkg-config: /usr/bin/pkg-config (0.29.2)
      Run-time dependency Boost found: YES 1.74.0 (/usr/include)
      Run-time dependency dl found: YES
      Run-time dependency threads found: YES
      Check usable header "libdivide.h" : YES
      Dependency boost (modules: program_options) skipped: feature tools disabled
      Dependency boost (modules: unit_test_framework) skipped: feature unit_test disabled
      Compiler for language cuda skipped: feature cuda disabled
      Run-time dependency libibverbs found: YES 1.14.47.0
      Run-time dependency librdmacm found: YES 1.3.47.0
      Run-time dependency libmlx5 found: YES 1.24.47.0
      Run-time dependency pcap found: YES 1.10.1
      Dependency libcap skipped: feature cap disabled
      Checking for function "ibv_modify_qp_rate_limit" with dependencies libibverbs, librdmacm: YES
      Checking for function "recvmmsg" : YES
      Checking for function "sendmmsg" : YES
      Checking for function "eventfd" : YES
      Checking for function "sem_init" with dependency threads: YES
      Checking for function "pthread_setaffinity_np" with dependency threads: YES
      Checking if "function multi-versioning" compiles: YES
      Checking if "MOVNTDQ intrinsic" compiles: YES
      Configuring common_features.h using configuration

      ../../src/meson.build:176:17: ERROR: Tried to form an absolute path to a dir in the source tree.
      You should not do that but use relative paths instead, for
      directories that are part of your project.

      To get include path to any directory relative to the current dir do

      incdir = include_directories(dirname)

      After this incdir will contain both the current source dir as well as the
      corresponding build dir. It can then be used in any subdirectory and
      Meson will take care of all the busywork to make paths work.

      Dirname can even be '.' to mark the current directory. Though you should
      remember that the current source and build directories are always
      put in the include directories by default so you only need to do
      include_directories('.') if you intend to use the result in a
      different subdirectory.

      Note that this error message can also be triggered by
      external dependencies being installed within your source
      tree - it's not recommended to do this.


      A full log can be found at /home/jsmith/spead2/build/cp310/meson-logs/meson-log.txt
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

```

I went to look at the line in question but I couldn't see reference to an absolute path there, unless the `pybind11.get_include()` function returns an absolute path?

bmerry:
> Everything looks fine, but I ran into an issue when I was following the instructions myself. I did it on dbelab07 just to ensure a fresh environment.

Did you perhaps put the virtual environment inside the source tree?

james-smith-za:
> Did you perhaps put the virtual environment inside the source tree?

What do you mean source tree exactly? The `src/` directory? I had a `.venv` directory in the root of the repo folder, the same as my convention on katgpucbf.

I tried deleting it and making a new one outside the repo folder entirely but ran into the same result.

bmerry:
> > Did you perhaps put the virtual environment inside the source tree?
> 
> What do you mean source tree exactly? The `src/` directory? I had a `.venv` directory in the root of the repo folder, the same as my convention on katgpucbf.

I mean the repo root. I can see how that would trigger the error about "Note that this error message can also be triggered by external dependencies being installed within your source tree - it's not recommended to do this." I'll update the document to indicate that this shouldn't be done.

> I tried deleting it and making a new one outside the repo folder entirely but ran into the same result.

Was it be any chance adjacent to the repository root and named spead2<something>? I've just done a test and it seems like there might be a bug with it using string startswith instead of a proper path check.

james-smith-za:
> I mean the repo root. I can see how that would trigger the error about "Note that this error message can also be triggered by external dependencies being installed within your source tree - it's not recommended to do this." I'll update the document to indicate that this shouldn't be done.

Noted!

> Was it be any chance adjacent to the repository root and named spead2? I've just done a test and it seems like there might be a bug with it using string startswith instead of a proper path check.

It was named `spead2-venv` so it would probably have triggerd that startswith bug.
I re-tried it quickly with a different name (`myvenv`) and that seems to have worked just fine!

james-smith-za (approved):

PR #266, opened 2023-09-05
Simplify wrapping of pointers to members

bmerry:
I did some microbenching and found no evidence that the SPEAD2_PTMF wrapper was improving performance at all, compared to just passing the pointer-to-member-function to pybind11, even though the latter involves making a call via a pointer (bad for branch prediction). So I've eliminated the SPEAD2_PTMF macro entirely, and simplified SPEAD2_PTMF_VOID to use a more obvious approach. The latter also isn't needed for any const member functions so I eliminated that support too.

### DIFF
--- a/include/spead2/py_common.h
+++ b/include/spead2/py_common.h
@@ -385,59 +385,21 @@ static inline U &&up(U &&u) { return std::forward<U>(u); }
 namespace detail
 {
 
-/* Some magic for defining the SPEAD2_PTMF macro, which wraps a pointer to
- * member function into a stateless class which avoids using a run-time
- * function pointer.
+/* Wrap a pointer-to-member to discard the return value. This is useful if it cannot
+ * be converted to a Python type.
  *
- * The type T and the Class template parameter are separated because if T
- * derives from Class and a member function foo is defined in Class, then the
- * type of &T::foo is actually <code>Return (Class::*)(Args...)</code> rather
- * than <code>Return (T::*)(Args...)</code>.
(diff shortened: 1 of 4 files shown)


bmerry:
@sjperkins I don't seem to be able to assign to you, but would you mind giving this a quick look?

include/spead2/py_common.h
```diff
@@ -385,59 +385,21 @@ static inline U &&up(U &&u) { return std::forward<U>(u); }
...
- * than <code>Return (T::*)(Args...)</code>.
+ * It's only implemented for non-const member functions, since const member
+ * functions generally only exist for the purpose of returning a result. The
+ * derived type needs to be explicitly supplied because pybind11 inspects the
+ * signature.
  */
-template<typename T, auto PTMF, typename Return, typename Class, typename... Args>
-auto make_ptmf(Return (Class::*)(Args...)) noexcept
+template<typename T, typename R, typename C, typename... Args>
+auto discard_result(R (C::*func)(Args...))
```
sjperkins, line 397: I think it would be helpful to retain the comment about `T` and `C` as separate but similar, because I had to think about it again and was helped by the removed the comment

sjperkins (changes requested):
I think it would be useful to retain some earlier comments, but otherwise it looks fine to me.

include/spead2/py_common.h
bmerry: Fair point. See if you're happy with the updated version.

bmerry:
Thanks for reviewing.

PR #282, opened 2023-09-21
Fix type annotations for send.UdpStream constructors

bmerry:
- Remove the old constructor which took a hostname and port (it was removed in spead2 4.0)
- Ensure that buffer_size is correctly handled for all the constructors that use it. This required duplicating each overload to handle two cases: if the following required arguments are passed positionally then config and buffer_size are required; if they are passed as kwargs then config and buffer_size are optional.


bmerry:
For reviewing: the overloads here need to match the actual implementations in `udp_stream_register` in py_send.cpp.

james-smith-za (approved):

PR #285, opened 2023-09-26
Allow stream to be explicitly started

bmerry:
Add stream configuration option to allow starting of data flow to be done explicitly by calling `start`. This allows multiple streams/readers to be started all at pretty much the same time, after the expensive part of creating the readers has been done.


james-smith-za (approved):
That was actually a simpler change than I'd anticipated based on what you'd described.

PR #289, opened 2023-10-24
Add AVX and AVX-512 non-temporal memcpy implementations

bmerry:
There are a few other pieces of restructuring:

- Instead of using the built-in function multi-versioning support, use an explicit ifunc. This has a few advantages:
  - The individual implementations can be called directly (used for unit testing).
  - The selection criterion can be customised (such as avoiding AVX-512 on Skylake)
  - It's easier to implement a fallback when the compiler/linker feature isn't available
- The actual implementation is in an internal header that is included multiple times, to avoid limitations in how GCC/clang handles target-specific functions.
- The unit tests now test all the implementations. They also cover a larger range of sizes (since the previous size was too small to fully exercise the AVX-512 code), but fewer alignments (to save time).


rtobar:
@bmerry can you quickly comment on the performance difference after these changes in selected avx512 platforms? Thanks!

bmerry:
> @bmerry can you quickly comment on the performance difference after these changes in selected avx512 platforms? Thanks!

The only AVX512 platform I have access to is Genoa (Zen 4). It makes a very minor difference (a percent or two), because it's bandwidth-limited. I'd be interested to see a comparison on recent Intel systems (if you have such systems, let me know - I can point you at some micro-benchmark code).

src/common_memcpy_impl.h
```diff
@@ -0,0 +1,113 @@
...
+ * options.
+ */
+
+namespace spead2
+{
+
+namespace detail
+{
+
+// U1 and U2 is unrolling factor; Ix is an index sequence 0, ..., Ux-1
```
ludwigschwardt, line 31: are unrolling factors?

ludwigschwardt (approved):
The bits I could understand looks good :-)

src/common_memcpy_impl.h
bmerry: Thanks, I started with just one factor and forgot to update the grammar.

PR #292, opened 2023-11-24
Update requirements

bmerry:


bmerry:
I think the MacOS wheel failures are because Homebrew just deleted the `boost@1.82` formula.

james-smith-za (approved):
Big bump in the Sphinx version.

I'm on a mobile at the moment but is all the documentation building fine?

bmerry:
> I'm on a mobile at the moment but is all the documentation building fine?

Seems to be okay on my machine, other than some deprecation warnings. I'll roll it back if readthedocs complains.

PR #294, opened 2023-11-28
Implement range-based for loop over ring_stream

bmerry:
Previously, the canonical way to iterate over all the received heaps in a ring_stream was:
```c++
while (true)
{
    try
    {
        auto heap = stream.pop();
        // Do stuff with heap
    }
    catch (spead2::ringbuffer_stopped &)
    {
        break;
    }
}
```

With this change, it's possible to rewrite that as
```c++
for (auto &&heap : stream)
{
    // Do stuff with heap
}
```

### DIFF
--- a/include/spead2/recv_ring_stream.h
+++ b/include/spead2/recv_ring_stream.h
@@ -190,13 +190,17 @@ bool ring_stream<Ringbuffer>::iterator::operator!=(const sentinel &) const
 template<typename Ringbuffer>
 auto ring_stream<Ringbuffer>::iterator::operator++() -> iterator &
 {
+    /* Clear it first, so that we can reclaim the memory before making
+     * space available in the ringbuffer, which might cause another
+     * thread to allocate more memory.
+     */
+    h = std::nullopt;
     try
     {
         h = stream.pop();
     }
     catch (ringbuffer_stopped &)
     {
-        h = std::nullopt;
     }
     return *this;
 }


sjperkins (approved):

bmerry:
@sjperkins thanks for the review. A slight improvement occurred to me in the meantime, which I've just pushed. Could you take a look? It's just a couple of lines.

include/spead2/recv_ring_stream.h
```diff
@@ -190,13 +190,17 @@ bool ring_stream<Ringbuffer>::iterator::operator!=(const sentinel &) const
 template<typename Ringbuffer>
 auto ring_stream<Ringbuffer>::iterator::operator++() -> iterator &
 {
+    /* Clear it first, so that we can reclaim the memory before making
+     * space available in the ringbuffer, which might cause another
+     * thread to allocate more memory.
+     */
+    h = std::nullopt;
```
sjperkins, line 197: How about setting this in the constructor instead?

sjperkins (changes requested):

include/spead2/recv_ring_stream.h
bmerry: I'm not following you. I want this to happen every time `operator++` is invoked, not just when the iterator is constructed. It already implicitly happens in the constructor, because that's the default-constructed state of a `std::optional<heap>`.

sjperkins: OK, I misunderstood then.

sjperkins (approved):

PR #302, opened 2024-01-23
Update GH CI to use upload/download-artifact@v4

bmerry:
This required some structural changes, because v4 no longer allows multiple matrix jobs to upload to the same artifact and have the results merged. So instead every job creates its own artifact and there is a final merge process (which also handles combining the debug symbols into a tarball).


james-smith-za (approved):
Looks fine.

PR #304, opened 2024-01-29
Support slice indexing on HeapReferenceList

bmerry:
This is a convenience for sending a partial list at the start or end of transmission. It's not optimal because it performs O(n) reference-counting operations (which is what HeapReferenceList was created to avoid), but that's no worse than Python list slicing. The alternative would be to build something akin to numpy's indexing, which just gives a view of the base object, but that would add a lot of complexity.


src/py_send.cpp
```diff
@@ -754,10 +776,10 @@ static void sync_stream_register(py::class_<T, stream> &stream_class)
     stream_class.def("send_heap", &T::send_heap,
                      "heap"_a, "cnt"_a = s_item_pointer_t(-1),
                      "substream_index"_a = std::size_t(0));
-    stream_class.def("send_heaps", &T::send_heaps,
-                     "heaps"_a, "mode"_a);
     stream_class.def("send_heaps", &T::send_heaps_hrl,
                      "heaps"_a, "mode"_a);
+    stream_class.def("send_heaps", &T::send_heaps,
+                     "heaps"_a, "mode"_a);
```
bmerry, line 782: The order had to be changed here because now that HeapReferenceList implements `__getitem__`, pybind11 thinks it is a Sequence and hence convertable to a `std::vector`. That causes it to be match T::send_heaps (which takes a vector). Pybind11 calls the first overload that matches, so it was trying to call the wrong overload.

james-smith-za (approved):
If you're going to use it only occasionally, i.e. at the beginning or end of transmission, then the performance impact seems not too big a deal?

bmerry:
> If you're going to use it only occasionally, i.e. at the beginning or end of transmission, then the performance impact seems not too big a deal?

Yes, that's the idea.

PR #311, opened 2024-02-08
Allow send rate to be overridden per-heap

bmerry:


include/spead2/send_stream.h
```diff
@@ -91,6 +97,11 @@ static inline std::size_t get_heap_substream_index(const heap &)
     return 0;
 }
 
+static inline double get_heap_rate(const heap &)
+{
+    return -1.0;
+}
+
```
james-smith-za, line 104: I'm not quite understanding the distinction between this one and the next one. The function is obviously overloaded. The `&` means that this version takes as its argument a reference to a heap. It's not named because it doesn't get used. I'm puzzled why this one always returns -1.

The one below takes a reference to a `heap_reference` - maybe it's my spead2 inexperience that's at play here but that's the behaviour I'd have expected.

james-smith-za:
As a more general query, I notice you don't include any testing in this PR.

Send rate management is probably a difficult thing to unit-test?

bmerry:
> Send rate management is probably a difficult thing to unit-test?

Indeed, it is particularly challenging when testing is on a cloud VM where you can't really rely on high-precision sleeps and reliable scheduling. I would love to be able to test this with mocking, but C++ isn't particularly conducive to that. I have done some manual testing.

include/spead2/send_stream.h
bmerry: A `heap_reference` bundles a heap with various options to override how it is sent; up to now that's just been heap_cnt and substream_index, and now includes transmission rate. If you pass a list of `heap` to `async_send_heaps`, you don't get to specify those things on a per-heap basis. If you want to specify any non-default values, you instead pass a list of `heap_reference`.

These get* functions provide abstraction so that the calling code doesn't need to know the difference between a `heap` and a `heap_reference`: it just passes whatever it has to (say) `get_heap_substream_index` to learn what substream index should be used.

james-smith-za: Ah, that does make sense.

james-smith-za (approved):

PR #312, opened 2024-02-09
Switch to netifaces2

bmerry:
netifaces is deprecated.

netifaces2 seems to have non-deterministic iteration order, which was causing the senders and receivers in TestPassthroughUdp6Multicast to use a different interface. Fix by adding @functools.lru_cache to cache the result after the first call.

### DIFF
--- a/tests/test_passthrough.py
+++ b/tests/test_passthrough.py
@@ -1,4 +1,4 @@
-# Copyright 2015, 2019-2023 National Research Foundation (SARAO)
+# Copyright 2015, 2019-2024 National Research Foundation (SARAO)
 #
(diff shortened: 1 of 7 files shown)


tests/test_passthrough.py
james-smith-za, line 1: The import doesn't change here - does netifaces2 call itself netifaces once it's installed?

bmerry: Yes. Which is actually pretty annoying, because it makes it impossible to have them both installed at once for a transition period.

james-smith-za: Not ideal but I guess if we must, we must.

james-smith-za (approved):

PR #315, opened 2024-02-15
Implement support for generic receive offload

bmerry:
GRO allows a socket to opt in to having the kernel group together packets with the same header information (source, destination, size etc) into one jumbo packet. The original packet size is provided as ancillary data to allow userspace to reconstruct the original packets. This can provide around a 20% speedup.

### DIFF
--- a/.ci/all-builds.sh
+++ b/.ci/all-builds.sh
@@ -24,4 +24,11 @@ for python in true false; do
             meson compile
         done
     done
+    # Similar recvmmsg and gro don't interact with either of the above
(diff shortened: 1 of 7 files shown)


bmerry:
When I ran some tests the results were mixed: with 1472-byte packets I got a 40% improvement, but with 8872-byte packets there was no difference. But it's also no worse, so probably worth having. I've also seen some comments on the internet that suggest that ConnectX-7 may have some hardware support for this.

james-smith-za (approved):

PR #317, opened 2024-04-18
Update CI builds on MacOS

bmerry:
- Run tests on macos-14 (Apple Silicon)
- Build x86_64 wheels on macos-13 and arm64 wheels on macos-14, so that
  you're doing native builds in each case. This simplifies the
  installation of libpcap since it's no longer necessary to cross-build
  it, and allows the arm64 wheels to be tested.
- Create a meson 'native' file which tells the compiler where to find
  Boost, includes and libraries on Homebrew. This is needed for Apple
  Silicon builds because Homebrew stuff goes into /opt/homebrew and the
  tools don't know about that.
- Update cibuildwheel to 2.17.0
- Unify some code between .ci/install-sys-pkgs.sh and
  .ci/before_all_macos.sh.

One regression with this change is that Python 3.8 wheels are no longer
produced for Apple Silicon. I'm guessing that this configuration is not
going to be in demand given that Apple Silicon support was first added
in Python 3.9 and only later backported to 3.8.

### DIFF
--- a/.ci/before_all_macos.sh
+++ b/.ci/before_all_macos.sh
@@ -4,35 +4,19 @@ set -e -u
 pcap_version=1.10.4
 
 # Prepare the MacOS environment for building spead2 wheels (for cibuildwheel)
-brew update
-brew install boost@1.84 libdivide
+.ci/install-sys-pkgs.sh
 
-# Install pkgconfig from source - once for each architecture in CIBW_ARCHS
+# Install libpcap from source
 cd /tmp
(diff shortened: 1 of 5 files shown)


ludwigschwardt:
Wow, those Linux aarch64s really take their time 😅

ludwigschwardt (approved):
Nice! I'm picking up notes for my own packages...

bmerry:
> Wow, those Linux aarch64s really take their time

Indeed. qemu is impressive, but it's not fast.

bmerry:
> Nice! I'm picking up notes for my own packages...

I figured you might want to review it for that reason.

PR #325, opened 2024-06-13
Unit tests: allocate ports dynamically

bmerry:
A number of tests uses hard-coded ports, which worked fine in Github Actions (each test runs in its own VM) but less well in the SARAO CBF Jenkins server (containerised, but uses host networking stack to allow the ibverbs tests to run). So rewrite things (in some cases with major refactoring) so that tests dynamically allocate ports.

There are still theoretical race conditions, because a "free" port is found by binding a socket to port 0, checking which port the kernel provided, then closing the socket. Between closing the ephemeral socket and opening the "real" one, another process may have grabbed that port. But it's much less racy than it was, where parallel execution had a very high chance of conflicts.


src/unittest_send_tcp.cpp
```diff
@@ -32,6 +32,18 @@ namespace spead2::unittest
...
+{
+    boost::asio::io_context context;
+    boost::asio::ip::tcp::socket socket(context);
+    socket.open(boost::asio::ip::tcp::v4());
+    socket.bind(boost::asio::ip::tcp::endpoint(boost::asio::ip::address_v4::loopback(), 0));
+    std::uint16_t port = socket.local_endpoint().port();
+    socket.close();
+    return port;
+}
+
```
james-smith-za, line 46: This is the potentially racy part I presume.

Perhaps I just lack the background to know what I don't know but can't you open the socket and then pass the socket along rather than the port number? That way would prevent another process from coming along and taking it.

james-smith-za (commented):
I don't really understand the inner workings of spead2 well enough to give this more than a cursory review.

All checks appear to have passed so I guess that's good?

src/unittest_send_tcp.cpp
bmerry: > can't you open the socket and then pass the socket along rather than the port number?

Yes, if the API you're testing works that way; but if you're testing an API that takes a port number then you're out of luck.

PR #357, opened 2024-11-04
Drop Python 3.8, which is End-of-Life

bmerry:
This simplifies the requirements files because numpy 2.0 spans Python
3.9-3.12. The versions in requirements.txt files are also bumped to the
latest since that was the easiest way to update them for Python 3.9 as
the minimum.


.pre-commit-config.yaml
```diff
@@ -35,36 +35,26 @@ repos:
...
-        files: '^(?:requirements\.in|requirements-3\.12\.txt)$'
       - name: pip-compile (py 3.13)
         id: pip-compile
         args: [--strip-extras, requirements.in, -o, requirements-3.13.txt]
         language_version: python3.13
         files: '^(?:requirements\.in|requirements-3\.13\.txt)$'
       - name: pip-compile (readthedocs)
         id: pip-compile
-        # --no-annotation because otherwise -r requirements-3.12.txt sometimes gets
+        # --no-annotation because otherwise -r requirements.txt sometimes gets
```
james-smith-za, line 56:
```suggestion
        # --no-annotate because otherwise -r requirements.txt sometimes gets
```
nit-picking, not that it makes a difference

james-smith-za (approved):
Are you not going to add something to the changelog?

bmerry:
> Are you not going to add something to the changelog?

I tend to do that when I'm ready to make a release, unless it's something I'm worried I'll forget to note.

.pre-commit-config.yaml
bmerry: Thanks for catching it.

PR #358, opened 2024-11-04
Bump versions of wheel dependencies

bmerry:


james-smith-za (approved):
Can't see a reason to object. Why was this necessary?

bmerry:
> Can't see a reason to object. Why was this necessary?

Not necessary, but since anyone installing a wheel gets stuck with whatever versions I bake into them, I like to keep them up to date.

PR #359, opened 2024-11-05
fixed the PDF version generation

shaoguangleo:
When I try to download the PDF version in [spead2](https://spead2.readthedocs.io/_/downloads/en/latest/pdf/),
It crashed. After checking the compile log, it looks like the error was caused by the SVG file.

```
! LaTeX Error: Cannot determine size of graphic in *.svg
```

This should fix the PDF generation host on the readthedocs.io, it works on my laptop at least.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -48,6 +48,7 @@ version, release = get_version()
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx.ext.imgconverter",
     "sphinx.ext.intersphinx",
     "sphinx.ext.mathjax",
     "sphinx.ext.napoleon",


bmerry:
@shaoguangleo thanks for raising this. I'm going to go with #360 instead though, which avoids rasterising the vector images.

shaoguangleo:
@bmerry Hi Bruce, thanks for your debug. The method metioned #359 worked on my MacBook, but the CI/CD failed.
I've already tried to fetch the branch you created #360 which was failed on my MacBook, but looks like the CI/CD works.
It looks weird, maybe caused by different PYTHON version. It's 3.10 on my laptop. 
I can not open the PDF on https://spead2.readthedocs.org/ right now, so maybe this will work after you merge the PR to the main branch.
So feel free to close this PR after #360 works.

> @shaoguangleo thanks for raising this. I'm going to go with #360 instead though, which avoids rasterising the vector images.

bmerry:
It looks like the PDF docs on readthedocs are working now, so I'm going to close this. If you notice any problems with the docs then feel free to open an issue.

PR #361, opened 2024-11-07
Add non-temporal memcpy support for ARM

bmerry:
It uses SVE, so it'll run on Neoverse CPUs like Grace, but doesn't support any Apple silicon. We're not using Apple for performance anyway (just to enable Apple users to develop) so at this stage I'm not proposing to implement alternative paths for non-SVE capable machines.


james-smith-za (approved):
I don't have any value to add here.

Sounds like a sensible idea to me.

PR #366, opened 2024-12-09
Jenkins: try to obtain cap_net_raw

bmerry:
setpriv isn't quite as convenient to use as spead2_net_raw (being more general purpose), but it has the advantage of already being installed and hence we don't need to mess around compiling the spead2 C++ tool suite.


james-smith-za (approved):

james-smith-za:
Not sure why we didn't run into this issue on Test Jenkins but I'm glad it's solved!

PR #373, opened 2025-02-10
Add annotation for ThreadPool.stop()

rtobar:
What it says on the tin


bmerry:
Thanks! Not sure how this slipped by for so long.

PR #374, opened 2025-02-12
Update versions used in testing

bmerry:
- Update all requirements*.txt files to use latest versions. In
  particular, numba now supports Python 3.13, which avoids the need
  for readthedocs to fall back to Python 3.12.
- Update minimum Ubuntu version to test to 22.04, because Github is
  retiring 20.04. This also means bumping the minimum Clang version to
  11.
- Use ubuntu-24.04 when testing latest versions of things, which allows
  to bump GCC to 14 and Clang to 18 (except coverage, because there seems to be an issue when using 24.04 for this)
- Test on 3.13 instead of 3.12 in general (previously 3.12 was used to
  cover the numba-related tests).

Also fix an outdated reference to Boost 1.69+; 1.70 is required.


requirements.txt
```diff
@@ -12,30 +12,30 @@ llvmlite==0.43.0
     # via numba
 netifaces2==0.0.22
     # via -r requirements.in
-numba==0.60.0 ; python_version < "3.13"
+numba==0.60.0
```
amishatishpatel, line 15: Why is the comment no longer necessary? The version stayed the same in this requirements file - I gather for python < 3.13.

amishatishpatel (approved):
One comment for clarification. Otherwise, good on you for navigating all these version updates in one go.

requirements.txt
bmerry: This is a file generated from requirements.in by pip-compile. Since requirements.in no longer declares `numba; python_version < "3.13"`, pip-compile no longer puts that in requirements.txt. requirements.txt is used for Python 3.9-3.12.

amishatishpatel: Ah, I thought you/a human added this comment.

PR #387, opened 2025-06-17
Add more pytest.mark.timeout

bmerry:
In particular, this ensures that testing the ibv support on a system without working multicast loopback will time out fairly quickly instead of hanging forever.

### DIFF
--- a/tests/test_passthrough.py
+++ b/tests/test_passthrough.py
@@ -904,6 +904,7 @@ class TestPassthrough:
         _test_item_group(transport_class, ig)
 
 
+@pytest.mark.timeout(5)
 @pytest.mark.transport(filter=lambda cls: cls.supports_substreams)
 class TestPassthroughSubstreams:
     """Tests for transports support multiple substreams."""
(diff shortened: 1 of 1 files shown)


tests/test_passthrough.py
```diff
@@ -966,6 +967,7 @@ def test_group_modes_mixed_sizes(self, transport_class, group_mode):
         _test_item_groups(transport_class, item_groups, group_mode=group_mode)
 
 
+@pytest.mark.time(5)
```
amishatishpatel, line 970: I'm spending more time setting up a local environment than intended. Basically,
- Why `pytest.mark.time` here and `.timeout` elsewhere?

amishatishpatel (commented):
One comment for your attention while I get my local env in order.

tests/test_passthrough.py
bmerry: Whoops, that's a typo! Thanks.

bmerry: That one doesn't actually need the timeout mark to fix the immediate issue, because it's not running any ibv tests, but it seemed harmless to sprinkle it on anyway. That's why testing didn't catch the typo (you can declare any mark you like, and it's up to plugins to decide what to do with them).

amishatishpatel (approved):
I was stressing because I could find documentation on `.mark.timeout` but none on `.time`!

bmerry:
> I was stressing because I could find documentation on `.mark.timeout` but none on `.time`!

Sometimes the problem is just between my keyboard and my chair :-)

